PR #3, opened 2011-04-06
Untitled

theleoborges:
Hey guys, I started having some issues with paranoia + dependent destroys and the culprit was the order in which the deleted_at attribute gets updated. I switched those around and it now behaves as expected in those scenarios.

Cheers,

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -1,8 +1,8 @@
 module Paranoia
   def destroy
-    self[:deleted_at] ||= Time.now
-    self.save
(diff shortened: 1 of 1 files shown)


radar:
Thank you for taking the time to patch this issue, it's much appreciated.

radar:
I've released the current version as 1.0.0 if you want to update your Gemfile to that now.

theleoborges:
No worries!

Tks for merging it and pushing the new version so fast! :)

Cheers

PR #5, opened 2011-05-12
Hi radar!

donv:
- Added query to ask a class or instance if it is paranoid
- Added test setup


radar:
Thanks for the patch! 1.0.1 has been pushed to Rubygems and it should include this patch.

donv:
Thank you!

PR #7, opened 2011-05-23
Preserve ActiveRecord semantics / Don't require models to be valid before destroy or delete.

tonybyrne:
Hi Radar,

Thanks for re-implementing a very useful plugin.

I've forked it to add some tweaks.  I noticed that trying to destroy / delete an invalid model failed.  Since we are blowing away the mode, I think it's better that this not be a restriction, so I've changed the implementation to skip validations on save.

Also, the original AR implementations return a frozen instance of the destroyed / deleted model instance. I've re-instated this behaviour in my fork.

Regards,

Tony.

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -10,8 +10,10 @@ module Paranoia
   def destroy
     _run_destroy_callbacks
     self[:deleted_at] ||= Time.now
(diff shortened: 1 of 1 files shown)


radar:
Good stuff, thanks for spending the time to work out how to fix it.

PR #8, opened 2011-05-24
Preserve ActiveRecord semantics / Don't require models to be valid before destroy or delete.

leh:
Hi Radar,

the in Issue #7 proposed (and pulled) changes don't fix the problem for me. I've extended the tests to provoke the problem (see fb9c903a8ff0097359cda631a410d867ae606aef).
For me c14aa4b6bd356ffe1cae48aee50b1f4c5e3f983c fixes those problems, so give it a try :-)

greetings
leh

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -9,11 +9,7 @@ module Paranoia
 
   def destroy
     _run_destroy_callbacks
-    self[:deleted_at] ||= Time.now
-    # If the instance has already been persisted, then we need to re-save it to flag it as
-    # destroyed / deleted.  We don't require validation in case it causes the updated to fail.
-    save(:validate => false) if persisted?
-    freeze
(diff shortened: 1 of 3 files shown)


radar:
I'm sorry I'm going to have to reject this, your commit c14aa4b bumps the activerecord version up to requiring >= 3.1.0.beta1, meaning people would have to install it to use this gem. Please fix this problem.

leh:
If you refer to line 18 of paranoia.gemspec in c14aa4b please notice that the line is commented out.
If line 18 isn't the problem, please give me a hint where to find this dependency, I have no clue how else I could have introduced that new dependency.

leh:
Cool, thx

radar:
Thanks for the commit man, sorry about incorrectly stating that AR version was bumped to 3.1

PR #9, opened 2011-05-26
Supporting a bit more of the acts_as_paranoid public API

asanghi:
@radar ,

Hey buddy! Not sure if the goal of the project is to remain lean or be a drop in replacement for the acts_as_paranoid, so I'm requesting you to consider supporting find_with_deleted. Please consider. Let me know if you want this modified in some way?

Added tests too!

Cheers

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -26,6 +26,10 @@ class ActiveRecord::Base
     default_scope :conditions => { :deleted_at => nil }
   end
 
+  def self.find_with_deleted
+    unscoped 
+  end
+
   def self.paranoid? ; false ; end
   def paranoid? ; self.class.paranoid? ; end
 end
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -5,6 +5,8 @@ module Paranoia
 
   module Query
     def paranoid? ; true ; end
+
+    def find_with_deleted ; unscoped ; end
   end
 
   def destroy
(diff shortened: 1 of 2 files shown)


radar:
I would rather people just used the `unscoped` method for this one. Although, if it's for backward compatibility, sure, why not? Can I ask why you didn't just alias the method?

asanghi:
Firstly, i think i added this at the wrong spot! We only want find_with_deleted when we include Paranoia. Will fix this.

I'm dyslexic in so far that i can never remember which method goes first in alias so i tend to overlook it. 
Same with using `ln -s` on unix. I'll try with alias.

Also in the same vein. I was going to send a pull for `find_only_deleted` but no more :)
We're transitioning an older app to Rails 3 and i'd rather use paranoia and hopefully make it a drop in as far as our API dependency on the older acts_as_paranoid.

asanghi:
i tried `alias :find_with_deleted :unscoped` but that doesn't work because `unscoped` is not a method on AR::Base.

asanghi:
Bah! The older API supported :all, :every and just one with `find_with_deleted` and `find_only_deleted`. 

So my pull isnt exactly backwards compatible. Sorry for the bother!

radar:
Perhaps you could add a note to the readme that if people want this functionality they could use `unscoped` and `unscoped.where("deleted_at IS NOT NULL")` respectively?

PR #12, opened 2011-06-13
Deletion and Restoration

efoxepstein:
`delete` is now a separate action from `destroy`. It skips callbacks and validations and whatnot, whereas `destroy` does not.

I also added `restore`, which silently updates `deleted_at` to be nil. It does _not_ unfreeze the attributes hash.

I added tests for all of this stuff.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -3,6 +3,6 @@ Bundler::GemHelper.install_tasks
 
 task :test do
   Dir['test/*_test.rb'].each do |testfile|
-      load testfile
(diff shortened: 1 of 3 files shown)


efoxepstein:
I also added actual nonparanoid deletion methods with bangs. I meant to put this on a feature branch but forgot.

radar:
Thank you, I have merged these changes. I removed some debug output with @3a58c9b. I am pushing 1.1.0 now with these changes.

Issue #17, opened 2012-01-26
Calling to_param on destroyed ActiveRecord object returns nil


curtis:
When creating an admin interface to restore destroyed objects, I ran across a problem when retrieving the path to the resource. Rails calls the to_param method during this process and it was returning nil for destroyed objects.

As you can see, the ActiveModel [to_param method](https://github.com/rails/rails/blob/master/activemodel/lib/active_model/conversion.rb#L49) calls `persisted?`, which in the ActiveRecord [persisted? method](https://github.com/rails/rails/blob/master/activerecord/lib/active_record/persistence.rb#L64) checks if it's a `new_record?` or `destroyed?`. Since Paranoia implements a `destroyed?` method, it will return true for paranoid destroyed objects.

A solution for this would be for paranoia to implement the ActiveRecord `persisted?` method which only checks for new_record if paranoid. For example:

```
def persisted?
  paranoid? ? !new_record? : super
end
```

PR #18, opened 2012-01-26
Implement persisted? method for ActiveRecord

curtis:
This should close issue #17


radar:
Could you please amend this commit (or create a new one) to not include files such as the .rvmrc? Ideally, this commit will _only_ be about implementing the `persisted?` method.

curtis:
Hi Ryan, how's that look?

radar:
Much better, thanks :)

PR #23, opened 2012-04-11
Fixes for Rails 3.2.* and some refactors

tiegz:
- use update_all instead of update_attribute in delete() and restore!() so update callbacks aren't run before/after deleting & restoring record
- update callback syntax

[edit 2/28/13: updated PR with pared down changes list]


lib/paranoia.rb
```diff
@@ -14,22 +14,32 @@ def only_deleted
...
   end
 
-  def destroyed?
+  # Has this record been marked as deleted? 
+  # This serves a different purpose than ActiveRecord#destroyed?, which checks 
+  # if the record in memory has been marked as destroyed.
+  def deleted?
     !self.deleted_at.nil?
   end
-  alias :deleted? :destroyed?
```
radar, line 34: Why did you remove this alias?

paranoia.gemspec
```diff
@@ -14,7 +14,7 @@ Gem::Specification.new do |s|
   s.required_rubygems_version = ">= 1.3.6"
   s.rubyforge_project         = "paranoia"
   
-  s.add_dependency "activerecord", ">= 3.0.0"
+  s.add_dependency "activerecord", "3.0.10"
```
radar: Why freeze this to a specific revision? Wouldn't `~> 3.0` be better?

test/paranoia_test.rb
```diff
@@ -98,14 +94,16 @@ def test_delete_behavior_for_callbacks
...
+    assert_equal nil, model.instance_variable_get(:@before_destroy_callback_called)
   end
   
   def test_destroy_behavior_for_callbacks
     model = CallbackModel.new
     model.save
     model.destroy
-    assert model.instance_variable_get(:@callback_called)
+    assert_equal nil, model.instance_variable_get(:@before_update_callback_called)
+    assert model.instance_variable_get(:@before_destroy_callback_called)
```
radar, line 175: Nice, I didn't know you could do this!

test/paranoia_test.rb
```diff
@@ -114,15 +112,18 @@ def test_restore
...
     
     model = ParanoidModel.only_deleted.find(id)
     model.restore!
     
+    assert_equal false, model.deleted?
     assert_equal false, model.destroyed?
   end
   
   def test_real_destroy
+    puts "Rails.version: #{ActiveRecord::VERSION::STRING}"
```
radar: You've left a `puts` statement here.

lib/paranoia.rb
tiegz, line 34: It mentions this in the pull request, but I did this for best practices. ActiveRecord defines this method to check if the in-memory record has been destroyed, similar to `new_record?` It checks the `@destroyed` ivar, which we can set to `true` and avoid having to rewrite the method. 

For instance, if I have an arbitrary ivar named `@some_record`, and I want to know if it's been destroyed in this request, I call `@some_record.destroyed?`. But if that class happens to have `acts_as_paranoid`, it might not have been destroyed in this request but would previously return `true`. Probably not a huge bug, but I felt it was important not to change the meaning of the method.

test/paranoia_test.rb
tiegz: thx! Will update shortly.

paranoia.gemspec
tiegz: Yeah, it was just in case there've been any big fixes in any tiny versions. I'll revert tho, no big deal.

tiegz:
hey @curtis , sorry for the delay, but I just pared down this pull request. Now it only replaces the usage of `update_attribute` with `update_all`, which should fix the clash with `acts_as_audited`.

tiegz:
@radar and @curtis : wondering if you guys have seen my changes in this pull request since yesterday?

curtis:
Hi @tiegz! Yes, I just had a chance to look it over. The changes look good to me.

tiegz:
Hey all, just realized this PR was still lying around. I think we pared it down to just a few tiny upgrades, and nothing too controversial. Should this be merged in now?

tiegz:
(no activity lately, closing this ticket)

radar:
@tiegz If you cared about having activity on this ticket, then the best thing to do would be to ping the maintainer by name.

lib/paranoia.rb
```diff
@@ -14,16 +14,24 @@ def only_deleted
...
 
   def destroy
-    _run_destroy_callbacks { delete }
+    run_callbacks(:destroy) { delete }
   end
 
   def delete    
-    self.update_attribute(:deleted_at, Time.now) if !deleted? && persisted?
+    if persisted?
+      self.deleted_at = Time.now
```
radar: I don't understand the need for this change to use `update_all`. Why?

lib/paranoia/version.rb
```diff
@@ -1,3 +1,3 @@
 module Paranoia
-  VERSION = "1.1.0"
+  VERSION = "1.2.1"
```
radar: Please don't change version numbers within PRs.

lib/paranoia.rb
radar: Ah, I just saw your comment here: https://github.com/radar/paranoia/pull/23#issuecomment-5691507.

Is that definitely still a problem?

radar:
Just a couple of things to clarify and then I think this is good to merge. Sorry that it slipped through the cracks.

lib/paranoia.rb
tiegz: @radar the latest version of paranoia uses `update_column` (down to rails 3.1), which I think should fix the clash with `acts_as_audited` (my reasoning was just to avoid callbacks, iirc)

tiegz:
@radar np, if possible I'll test the latest version with `acts_as_audited` tmw and see if the other changes here (eg `if persisted?`) are still necessary

radar:
@tiegz Great. Let me know how it goes! If I don't reply by the end of the week then please buzz me again.

tiegz:
@radar just merged master into this branch (opting for the newer preference for `update_column` from master). Looks like the only remaining changes would be the callback avoidance tests, which look to be covered in master anyway (https://github.com/radar/paranoia/commit/d28a4a0313ec9cedb7b02671a1f47318b5b2681e)

Cool if we close this PR out now?

radar:
@tiegz Seems like a good idea to me. Thanks :)

PR #25, opened 2012-06-26
Update_attribute is gone in Rails 4, add update_column support.

erichmenge:
update_attribute is deprecated in Rails > 3.2.6. Use update_column if available.


radar:
:heart:

PR #27, opened 2012-06-28
Fix for ambiguous column 'deleted_at' issue

defconomicron:
Fixes possible ambiguous column 'deleted_at' issue if the following were attempted:
Alpha.joins(:beta).only_deleted


thromera:
I still have the issue, but now on 'active' column (because our model has an active + deleted_at column). Is it fixed ? (Sorry to bump 4 years)

PR #32, opened 2012-10-22
paranoia should work well with associations, unscoped is evil

joshgoebel:


stellard:
+1 

thanks for this! works great.

joshgoebel:
Radar, can we get this merged?

PR #35, opened 2012-11-21
Add support for dependent: :destroy! and :delete!

arsduo:
We've been using Paranoia in one of our projects at @6Wunderkinder, and it's been extremely useful.  However, one gap we've found is that there are cases when you want to hard-destroy dependent resources -- for instance, when a user destroys her account, all related resources should be wiped from the database rather than left there orphaned.

This pull request adds support for `dependent: destroy!` (for HasOne and HasMany) and `dependent: delete!` (for HasOne -- delete_all for HasMany works out of the box).  These options function as expected, calling callbacks or not and then hard-deleting the record.  

This works for Rails 3.1 and 3.2 (tested with 3.1.8, 3.2.0, and 3.2.9).  I started doing work on 3.0.x, but it turns out that the current version of Paranoia doesn't work with 3.0.x anyway -- ditto 4.0, which I may take a crack at later.  (It's unfortunately involves copying some code from ActiveRecord, but it's relatively minimal (and should be less once paranoia is extended to cover Rails 4.0.)

Thanks for the great gem!

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ class AddDeletedAtToClient < ActiveRecord::Migration
   end
 end
 ```
-    
+
 ### Usage
 
 #### In your model:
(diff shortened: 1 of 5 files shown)


radar:
I'm :-1: on this _because_ it hacks at internal Rails API. I wish there was a way to do this without hacking at Rails.

arsduo:
Turns out there is!  I was thinking about it more after reading your comment (which is, by the way, entirely correct), and realized that we can implement this using an after_destroy callback.  It's a bit more cumbersome, but it does the job.  Closing this and submitting a new pull-request now.

PR #36, opened 2012-11-22
Support for dependent: :destroy! and :delete!, take 2

arsduo:
This pull request reimplements #35 without mucking around in Rails internals.

There are some cases where you want to hard-destroy dependent resources that might otherwise be paranoid -- for instance, when a user destroys her account, all related resources should be wiped from the database rather than left there orphaned.

This pull request adds support for such dependent hard-deletion using two new class methods:

``` ruby
class User
  # these methods are available to all classes, 
  # since non-paranoid classes may have paranoid dependencies
  hard_delete_dependencies :widgets, :account
  hard_destroy_dependencies :schematics
end
```

When a user is destroyed, the associated resources will be destroyed or deleted (with or without callbacks, respectively) as expected.

A few points of note:
- Since this uses an after_destroy callback, it only works when the resource is destroyed, not deleted (unlike AR's dependent options).  I think this is an acceptable cost, and is documented in the readme.
- For the same reason, you need to call hard_*_dependencies after any other after_destroy callbacks that depend on the children being present; this is also documented.
- This should work on all modern versions of Rails, since it only relies on after_destroy and find_in_batches.

The changes are tested.

Thanks!


arsduo:
I've just realized that this is fine for non-paranoid classes, which are only have one type of deletion (hard); for paranoid classes, this should be smart enough to differentiate the model being soft-destroyed (in which case it would soft-delete the dependent resources) and hard-destroyed, in which case it should kill related records dead.  

That should be straightforward to implement if there's a clean way to see whether a record was hard- or soft-deleted in the after_destroy callbacks; if we kept a reference to the original :deleted? method before overwriting it with :destroyed?, that should work.

lib/paranoia/active_record.rb
```diff
@@ -0,0 +1,111 @@
...
+    def self.hard_delete_dependencies(*dependencies)
+      hard_remove_dependencies(:delete!, dependencies)
+    end
+
+    # Internal: Relations that have been marked for hard-destroy (with callbacks).
+    def self.hard_destroy_relations
+      @_hard_destroy_relations ||= []
+    end
+
+    # Internal: Relations that have been marked for hard-destroy (with callbacks).
```
radar: Comment here is a dupe of the above one. I think you meant "hard-delete (without callbacks)"

arsduo: Fixed!  Good catch.  (Sorry for the delay, was out of town on vacation.)

arsduo:
Hmm, suddenly a Travis failure.  I'll look at that.

radar:
The Travis failure is because of the restore test, which is unrelated to
your changes... but if you want to fix it, I won't stop you.

Issue #37, opened 2012-11-28
.restore! does not have any effect


vbrendel:
With version 1.2.0, calling Thing.only_deleted.find(123456).restore! results in no effect, the SQL that gets run is:

UPDATE "things" SET "deleted_at" = NULL WHERE "things"."deleted_at" IS NULL AND "things"."id" = 123456

It probably has to do with the fixes to the scoping because it worked in previous version.

radar:
The [test for restore](https://github.com/radar/paranoia/blob/cb018cc361e44ed2f0c774749f5e0429f4757a7d/test/paranoia_test.rb#L178-L190) actually tests this exact case. The build passes here.

Do you have some steps to reproduce this issue on a fresh app?

vbrendel:
Hi @radar,

The test passes indeed because destroyed? returns false immediately after calling restore! even though it didn't delete the row from the database. In the test, if you reload the model after doing restore then destroyed? returns true again.

So I would propose to rewrite the test like this:

``` ruby
  def test_restore
    model = ParanoidModel.new
    model.save
    id = model.id
    model.destroy

    assert model.destroyed?

    model = ParanoidModel.only_deleted.find(id)
    model.restore!

    model = ParanoidModel.find(id)

    assert_equal false, model.destroyed?
  end
```

radar:
I can see now.

It doesn't look like anything that we're doing. I think it _might_ be an Active Record bug? I am not sure how to investigate this further.

radar:
I've broken the test for the `restore!` method on purpose now.

linjunpop:
@radar It is a ActiveRecord bug. https://github.com/rails/rails/issues/7357

linjunpop:
A workaround in my application is in an initializer:

``` ruby
module Paranoia
  # override
  def restore!
    true if self.class.unscoped.update_all({deleted_at: nil}, self.class.primary_key => id) > 0
  end
end
```

https://github.com/rails/rails/commit/a3c3cfd

radar:
Thanks for the heads up!

linjunpop:
Rails 3.2.13 should works.

PR #39, opened 2012-11-30
Change that addresses #37

vbrendel:
update_attribute does not seem to have this problem, where update_column does. So I'm proposing to use update_attribute where available.

I'm really not sure about older versions of Rails however so it may need more testing.


lunks:
Looking at #37, it looks like this issue can be closed now.

radar:
Thanks @lunks for the heads up.

PR #42, opened 2013-02-25
fix alias methods

zhengjia:
The test `test_real_destroy` failed when the test is run with ruby 2.0
The fix uses `alias` instead of `alias_method` to make sure `destroy` in active_record is used.


radar:
Thanks!

PR #43, opened 2013-02-27
Adding Gemfile.lock to .gitignore. According to [Yehuda's post](http://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/):

lunks:
```
When developing a gem, use the gemspec method in your Gemfile to avoid duplication. In general, a gem’s Gemfile should contain the Rubygems source and a single gemspec line. Do not check your Gemfile.lock into version control, since it enforces precision that does not exist in the gem command, which is used to install gems in practice. Even if the precision could be enforced, you wouldn’t want it, since it would prevent people from using your library with versions of its dependencies that are different from the ones you used to develop the gem.
```


linjunpop:
:+1:

radar:
Seems good to me. Thanks @lunks.

lunks:
Thanks for merging!

PR #44, opened 2013-02-27
Adding test with joining a paranoia model.

lunks:
This is something I've stumbled on and lost a few hours. If I remove a child object and then query for parent objects which relates to child models, I find records I should not.

Merging {deleted_at: nil} on the child query works.

While this might be a bug on Rails, or even might be intended behavior, I think the default scope should apply here. What do you think?

### DIFF
--- a/test/paranoia_test.rb
+++ b/test/paranoia_test.rb
@@ -73,7 +73,7 @@ class ParanoiaTest < Test::Unit::TestCase
     assert_equal 0, model.class.count
     assert_equal 1, model.class.unscoped.count
   end
-  
+
   def test_scoping_behavior_for_paranoid_models
     ParanoidModel.unscoped.delete_all
     parent1 = ParentModel.create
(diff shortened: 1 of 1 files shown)


lunks:
Any updates on this?

radar:
I think that the default scope should apply too. Could you please investigate a patch for this issue?

bteitelb:
This bit me too.  The proposed solution over at goncalossilva/acts_as_paranoid is to override `ActiveRecord::Associations::JoinDependency::JoinAssociation#build_constraint`: https://github.com/zlx/acts_as_paranoid/commit/6bb9292c3f79483a014186cf5ded6698c657262e

bteitelb:
I tried a different approach, which involved adding a condition to any association with a paranoid model https://github.com/bteitelb/paranoia/commit/4745fb4b83f1f3238d5e5f22f49fc6b9550e69d5 :

``` ruby
self.reflect_on_all_associations.each do |reflection|
  if (klazz = reflection.klass).paranoid?
    options = reflection.instance_variable_get(:@options)
    options.deep_merge!({:conditions => klazz.where(klazz.quoted_table_name + ".#{klazz.paranoia_column} IS NULL")})
    reflection.instance_variable_set(:@options, options)
  end
end
```

From the console, I was able to use a similar approach to add a condition to the SQL that's generated by joining through an association.  This code, however, did not work.  I'm not sure why.

radar:
@bteitelb Would you please be able to investigate a patch for this issue?

PR #46, opened 2013-03-16
Fix compability with rails 4.0

andryunin:
ActiveSupport callbacks changed in Rails 4.0: now it does not create _run_*_callbacks methods.


radar:
Thanks @andryunin!

PR #50, opened 2013-04-11
Documentation clarification

thomasv314:
Howdy,

I appreciate this gem alot. I hadn't used the old acts_as_paranoid, so when I first wrote my tests they were failing and I was confused as to why...

Figured this commit might save another newbie a few minutes...

Thanks,

Thomas V.


radar:
Looks good, thanks @tommyvyo :)

PR #51, opened 2013-04-14
More info about imgration on readme

maurogeorge:
Explaining how to run migrations on rails 3 and 2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,18 +46,22 @@ rake gems:install
 
 #### Run your migrations for the desired models
 
-```ruby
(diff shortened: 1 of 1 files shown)


radar:
Paranoia does not support Rails 2. Could you please remove that information from the README? Other than that, this patch looks fine.

maurogeorge:
Remove any reference to rails 2 as you can [see in here](https://github.com/maurogeorge/paranoia/tree/readme-migration#run-your-migrations-for-the-desired-models).

PR #53, opened 2013-04-22
Fixes ActiveRecord 4.0 warning

zimbatm:
It seems that AR 4.0 has changed how to define default scopes.


radar:
Seems good. Thanks @zimbatm!

PR #56, opened 2013-05-17
use Model.all instead of Model.scoped

delba:
Since Rails 4, Model.all returns an ActiveRecord::Relation, rather than an array of records. 
This PR gets rid of the deprecation warnings BUT it will break some code.


inkstak:
[Please excuse this message, I've send it to the wrong repository]

radar:
This will break Rails 3 compatibility if we merge it to master. Will create a rails4 branch.

radar:
Added :)

PR #60, opened 2013-05-17
use a GuardClause

delba:
Using a GuardClause to check if the record is a `new_record?` or if it has been `destroyed?` reads better IMHO.
Also, using `destroyed?` rather than `deleted?` because the latter is just an alias.


radar:
Looks good. Thanks @delba!

PR #71, opened 2013-10-04
destroy! when destroying deleted? records

michaltn:
It would be really nice if destroying a destroyed object deleted it from a db.

It provides for destroy working as expected for has_many relationships (first destroy marks them all, second call will remove them all)

It also allows for destroy_all to work for destroying records permanently with 'with_deleted.destroy_all'

This is consistent with other act_as_paranoid gems


radar:
Hi @michaltn, thanks for the patch. Could you please add a note to the README about this, as well as writing a test for this new behaviour?

Thanks!

michaltn:
Done :)

radar:
Thank you! :)

parkr:
:heart:

Issue #67, opened 2013-07-30
destroy! implemented in Rails 4


garysweaver:
Understand that paranoia is currently stated as a Rails 3 solution, but for future Rails 4 support:

https://github.com/rails/rails/blob/v4.0.0/activerecord/lib/active_record/persistence.rb#L169

Maybe force_\* methods would be better than bang methods? (e.g. force_destroy instead of destroy! and force_delete instead of delete!)

radar:
@garysweaver There is a rails4 branch of this extension. Have you tried that yet? I like the idea of `force_` methods.

MarkMurphy:
@radar why can't we just do destroy!(true) for forced deletes?

radar:
@MarkMurphy Would that work? I'd like a patch for that if it does.

radar:
@MarkMurphy ping

MarkMurphy:
@radar Check out my fork of [paranoia](https://github.com/MarkMurphy/paranoia/tree/master) which could be made into a pull request.

radar:
@MarkMurphy Please do submit a PR so I can look it over.

jhawthorn:
Closing because we now have `really_destroy!`?

PR #75, opened 2013-10-24
use force_* methods instead of bang methods

delba:
It should solve #67.

However, **this is not backward compatible** as the bang methods (`destroy!` and `delete!`) have been completely removed.

Alternatively, we could keep the bang methods for now and add a deprecation warning (?)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,4 +1,4 @@
-source :gemcutter
+source 'https://rubygems.org'
 
 # Specify your gem's dependencies in paranoia.gemspec
 gemspec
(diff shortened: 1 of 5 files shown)


radar:
Hi @delba, this PR does not merge correctly according to GH. Could you please fix that up?

I am all for adding a deprecation warning for those methods and keeping it in there for a month or two. After that, we can review and remove that warning and do another release.

delba:
Ok, I just did something stupid. I rebased the original commit and then force pushed.
I will mess everything up so I guess I have to close this PR.

PR #76, opened 2013-10-24
Add configuration option for column name in Rails 4

quainjn:
Pulling over a feature from acts_as_paranoid gem 

https://github.com/goncalossilva/acts_as_paranoid


radar:
Looks great to me. Thanks! Added to the rails4 and master branches.

PR #78, opened 2013-10-29
Make default_scope optional

joshuap:
This allows you to turn off default scoping with the following option:

`acts_as_paranoid default_scope: false`

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -11,6 +11,11 @@ module Paranoia
       all.tap { |x| x.default_scoped = false }
     end
 
+    def without_deleted
+      all.where paranoia_column => nil
+    end
+    alias :undeleted :without_deleted
+
     def only_deleted
       with_deleted.where.not(paranoia_column => nil)
     end
@@ -70,7 +75,7 @@ class ActiveRecord::Base
     class_attribute :paranoia_column
 
     self.paranoia_column = options[:column] || :deleted_at
-    default_scope { where(self.quoted_table_name + ".#{paranoia_column} IS NULL") }
+    default_scope { without_deleted } unless options[:default_scope] == false
   end
 
   def self.paranoid? ; false ; end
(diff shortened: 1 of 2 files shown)


brantwedel:
+1 multi-tenancy Rails 4 app here as well

I was experimenting with the `with_deleted` method, but no luck yet as `default_scope` gets applied by active record at the latest possible moment, I finally ended up with a hack that works but opens up potential problems.

The following works by applying all default scopes manually then filtering out the deleted_at scope but **please don't use it in production**:

``` ruby
    def with_deleted
      all.tap do |x|
        if x.default_scoped
          x.merge!(x.default_scopes.inject(x) do |default_scope, scope|
            if !scope.is_a?(ActiveRecord::Relation) && scope.respond_to?(:call)
              default_scope.merge(scope.call)
            else
              default_scope.merge(scope)
            end
          end)
          x.default_scoped = false
          x.where_values.delete_if { |query| (query.is_a?(String) && query.include?(".#{paranoia_column} IS NULL")) || (!query.is_a?(String) && query.to_sql.include?(".#{paranoia_column} IS NULL")) }
        end
      end
    end
```

radar:
@onecsguy Can you show me this default scope code for your multitenancy please?

brantwedel:
@radar 

```
class Contact < ActiveRecord::Base
  # Only access records from current SubmarineAccount using default scope, SubmarineAccount.current_account_id is set in the session to be the account of the currently logged in user
  default_scope { where(:submarine_account_id => SubmarineAccount.current_account_id) }
  acts_as_paranoid
end
```

My project began as an internal project management app, then I got the 'bright' idea to use `default_scope` to convert the whole thing to multi-tenancy with just 6 lines of code (well 21 when I upgraded to rails 4 and this gem).

I am basically using https://github.com/railscasts/388-multitenancy-with-scopes

radar:
@onecsguy Oh, right then :) I thought that that way of multitenancy wasn't thread-safe. I don't have any citations of that, so it's just like my opinion and stuff.

Speaking of opinions, did you know I have written a book about [Multitenancy with Rails](https://leanpub.com/multi-tenancy-rails)?

PR #79, opened 2013-10-29
Make default_scope optional

42gb:
I added the opportunity to make the default_scope optional. This is inspired by the commit of the fork of joshuap in the rails4 branch. (https://github.com/joshuap/paranoia/commit/ff15d315e0ec425b4e4183742485abe53cd22ef1).

I would be glad if this is going into the master branch of radar/paranoia. I hope this helps the community.

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -14,6 +14,11 @@ module Paranoia
       scoped.tap { |x| x.default_scoped = false }
     end
 
+    def without_deleted
+      where(paranoia_column => nil)
+    end
+    alias :undeleted :without_deleted
+
     def only_deleted
       with_deleted.where("#{self.table_name}.#{paranoia_column} IS NOT NULL")
     end
@@ -84,7 +89,8 @@ class ActiveRecord::Base
     class_attribute :paranoia_column
 
     self.paranoia_column = options[:column] || :deleted_at
-    default_scope { where(self.quoted_table_name + ".#{paranoia_column} IS NULL") }
+
+    default_scope { without_deleted } unless options[:default_scope] == false
   end
 
   def self.paranoid?
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -15,12 +15,12 @@ module Paranoia
     end
 
     def without_deleted
-      where(paranoia_column => nil)
+      where("#{self.quoted_table_name}.#{paranoia_column} IS NULL")
     end
     alias :undeleted :without_deleted
 
     def only_deleted
-      with_deleted.where("#{self.table_name}.#{paranoia_column} IS NOT NULL")
+      with_deleted.where("#{self.quoted_table_name}.#{paranoia_column} IS NOT NULL")
     end
     alias :deleted :only_deleted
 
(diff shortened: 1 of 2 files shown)


42gb:
@radar Is it possible that that pull request is going into master?

radar:
@42gb have some patience, please. I do more than just maintain this project :)

42gb:
@radar Sorry for that. I read in another PR of this project that it's a good thing to ping you. Patience mode on ;-).

radar:
What is the usecase for this? Why would people want this feature?

42gb:
Perhaps I didn't get the concept of this gem. Please correct me, if I'm wrong. I use it in a simple has_one connection between an order model and an user model. An order belongs to one user.

With the added scope of paranoia the view of order#show throws an error if the user of this order was deleted, because @order.user == nil. This is not what i wanted to achieve. I don't want to crash my views for orders, if an user is soft deleted and I also don't want to rewrite every controller action where I have a association with an user, which could be soft deleted. With not adding the standard scope to the users model with the newly added option, my orders view is still working. And I get the information connected to that user. My views stay untouched. 

Is there another way to achieve this?

I also did this PR because i saw that @joshuap added the same logic to the rails 4 branch. Perhaps he should join this discussion too. 

Cheers!

joshuap:
My use case is that I have a model which exists as part of a "directory" but is also used to maintain associations elsewhere in the application. I want to be able to "delete members from the directory", but the record should otherwise remain intact. Since I only need to scope to `undeleted` when displaying/searching the directory members, a `default_scope` would require me to use `unscoped` everywhere else.

raszi:
:+1: on this PR to be merged

radar:
@42gb I don't see how `@order.user` could be nil if it was the ORDER that was deleted. Please show me an application where this error happens.

42gb:
@radar "if the user of this order was deleted" -> not the order is deleted, the user assigned to this order is nil.

radar:
We work around this problem in Spree by [re-defining the association getter methods](https://github.com/spree/spree/blob/ef871a7e938c841ccbcbac7fcacad2efcf355929/core/app/models/spree/line_item.rb#L94-L96):

``` ruby
 def variant
   Spree::Variant.unscoped { super }
 end
```

I think having the default behaviour hide all records which are deleted and to have association overrides like the above is more sane than allowing a configuration option to disable it.

42gb:
@radar Thanks for that advice. This also helps a lot. Cheers!

dmitry:
@radar Don't you know how it's may be possible to do it better? Because in that case, some default_scopes that are not related to the paranoia gem (`deleted_at: nil`) may be cleaned out.

radar:
@dmitry I don't understand what you're saying. Show me an example of how it can be done better.

dmitry:
@radar my idea is similar to the unscoped method:

``` ruby
Spree::Variant.without_delete { Spree::Variant.all } 
```

but the difference is `without_delete` should remove only the `deleted_at IS NOT NULL` part from the WHERE of the scope. It can be two variants of doing the same thing, with block and with scope/method, like `Product.first.variant.without_delete`, I think it's possible after extension of the ActiveRecord's `Associations` class.

But for now, as I understand, `unscoped` with block is the only way of doing it.

PR #80, opened 2013-10-30
callback bugs

alfa-jpn:
Hello,

In rails, after_commit must be called, when destroy.

but, Paranoid doesn't call it, because it uses update_column.

I have solved this problem using update_attributes.

update_attributes (alias to update)
 deprecated : false
 callbacks  : true

update_attributes 
 deprecated : true
 callbacks  : true

update_column
 deprecated : false
 callback   : false

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -49,17 +49,26 @@ module Paranoia
 
   def delete
     return if new_record?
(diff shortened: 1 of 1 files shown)


alfa-jpn:
created update_attrs_without_valid function because update_attributes executes a validation.

PR #81, opened 2013-10-30
callback bugs for master

alfa-jpn:
Hello,

In rails, after_commit must be called, when destroy.

but, Paranoid doesn't call it, because it uses update_column.

I have solved this problem using update_attributes.

update_attributes (alias to update)
 deprecated : false
 callbacks  : true

update_attributes 
 deprecated : true
 callbacks  : true

update_column
 deprecated : false
 callback   : false

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -52,11 +52,11 @@ module Paranoia
 
   def delete
     return if new_record?
-    destroyed? ? destroy! : update_attributes(paranoia_column => Time.now)
+    destroyed? ? destroy! : update_attrs_without_valid(paranoia_column => Time.now)
   end
 
   def restore!
-    run_callbacks(:restore) { update_attributes(paranoia_column => nil) }
+    run_callbacks(:restore) { update_attrs_without_valid(paranoia_column => nil) }
   end
   alias :restore :restore!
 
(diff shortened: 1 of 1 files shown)


alfa-jpn:
created update_attrs_without_valid function because update_attributes executes a validation.

radar:
Why must it be called?

alfa-jpn:
It was stipulated in the Ruby on Rails documents that after_commit is called after a record has been created, updated, or destroyed.
http://api.rubyonrails.org/classes/ActiveRecord/Transactions/ClassMethods.html#method-i-after_commit

This document was also the same.
http://api.rubyonrails.org/classes/ActiveRecord/Callbacks.html

So, developers expect that after_commit be called after destroyed.

I think the callback is necessary to use paranoia without bug fixes.

lib/paranoia.rb
```diff
@@ -68,12 +68,13 @@ def destroyed?
...
   private
 
-  # Rails 3.1 adds update_column. Rails > 3.2.6 deprecates update_attribute, gone in Rails 4.
-  def update_attribute_or_column(*args)
-    self.class.unscoped do
-      respond_to?(:update_column) ? update_column(*args) : update_attribute(*args)
-    end
+  # update attributes without validation
+  # @param attributes [Hash] attributes
+  def update_attrs_without_valid(attributes)
```
radar, line 73: This code will still trigger normal update callbacks to happen on the model, which is not good. 

:-1: for this change.

radar:
I have added a regression test for this PR now to the master and rails4 branches. If you can make the code behave as you intend and still have the tests pass, then I would be welcome to that patch.

PR #82, opened 2013-11-02
Failing test for joining a table multiple times

johanneswuerbach:
With activerecord 4.0.1 just using the `quoted_table_name` doesn't work in all cases as tables are now aliased when joined multiple times. (jobs_employees below)

``` sql
SELECT COUNT(*) FROM "employees"
INNER JOIN "jobs" "jobs_employees" ON 
  "jobs_employees"."employee_id" = "employees"."id" AND ("jobs".deleted_at IS NULL)
INNER JOIN "jobs" ON
  "employees"."id" = "jobs"."employee_id"
WHERE
  ("employees".deleted_at IS NULL) 
  AND ("jobs".deleted_at IS NULL) AND "jobs"."employer_id" = $1
```

sqlite3 ignores the wrong reference, but postgresql doesn't

```
ActiveRecord::StatementInvalid: PG::UndefinedTable: ERROR:
  invalid reference to FROM-clause entry for table "jobs"
LINE 1: ..._employees"."employee_id" = "employees"."id" AND ("jobs".del...
                                                             ^
HINT:  Perhaps you meant to reference the table alias "jobs_employees".
```

I had to replace your plain sql model creation with ActiveRecord::Schema as PG is a bit different. Would you like to keep that?

### DIFF
--- a/paranoia.gemspec
+++ b/paranoia.gemspec
@@ -18,6 +18,7 @@ Gem::Specification.new do |s|
 
   s.add_development_dependency "bundler", ">= 1.0.0"
   s.add_development_dependency "sqlite3"
+  s.add_development_dependency "pg"
   s.add_development_dependency "rake", "0.8.7"
 
   s.files        = `git ls-files`.split("\n")
--- a/test/paranoia_test.rb
+++ b/test/paranoia_test.rb
@@ -2,22 +2,57 @@ require 'test/unit'
 require 'active_record'
 require File.expand_path(File.dirname(__FILE__) + "/../lib/paranoia")
 
-DB_FILE = 'tmp/test_db'
+# DB_FILE = 'tmp/test_db'
 
-FileUtils.mkdir_p File.dirname(DB_FILE)
-FileUtils.rm_f DB_FILE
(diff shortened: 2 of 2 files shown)


radar:
> I had to replace your plain sql model creation with ActiveRecord::Schema as PG is a bit different. Would you like to keep that?

Could you explain how it was different please?

johanneswuerbach:
@radar I was referring to the following sql lines: https://github.com/radar/paranoia/blob/rails4/test/paranoia_test.rb#L11

Postgres uses the `timestamp` data type instead of `DateTime` and requires to setup a sequence for auto incrementing ids. ActiveRecord::Schema handles those adapter specific things and simplifies testing with different database adapters.

I think the issue looks like a regression in activerecord 4.0.1 as all tests are passing with 4.0.0.

johanneswuerbach:
I created an issue for rails: rails/rails#12770

radar:
Thanks. I'll keep on an eye on that Rails issue now.

johanneswuerbach:
Fixed in rails 4.0.6.

radar:
Thank you @johanneswuerbach!

PR #83, opened 2013-11-02
Replaced deprecated gemcutter source

johanneswuerbach:
Changed source to remove deprecation warning while using bundler

```
The source :gemcutter is deprecated because HTTP requests are insecure.
Please change your source to 'https://rubygems.org' if possible, or 'http://rubygems.org' if not.
```


radar:
Thanks @johanneswuerbach!

PR #84, opened 2013-11-06
bugfix the callback behaviors.

alfa-jpn:
bugfix the callback behaviors. #81

and add tests for plain models callback.

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -77,9 +77,9 @@ module Paranoia
   # insert time to paranoia column.
   # @param with_transaction [Boolean] exec with ActiveRecord Transactions.
   def touch_paranoia_column(with_transaction=false)
-    return touch(paranoia_column) unless with_transaction
-
-    with_transaction_returning_status do
+    if with_transaction
+      with_transaction_returning_status { touch(paranoia_column) }
+    else
       touch(paranoia_column)
     end
   end
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -47,12 +47,12 @@ module Paranoia
   end
 
   def destroy
-    run_callbacks(:destroy) { delete }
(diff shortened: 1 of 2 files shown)


lib/paranoia.rb
```diff
@@ -67,11 +67,20 @@ def destroyed?
...
+  # insert time to paranoia column.
+  # @param with_transaction [Boolean] exec with ActiveRecord Transactions.
+  def touch_paranoia_column(with_transaction=false)
+    return touch(paranoia_column) unless with_transaction
 
-  # Rails 3.1 adds update_column. Rails > 3.2.6 deprecates update_attribute, gone in Rails 4.
-  def update_attribute_or_column(*args)
-    self.class.unscoped do
-      respond_to?(:update_column) ? update_column(*args) : update_attribute(*args)
+    with_transaction_returning_status do
```
radar: This looks to me like it would call `touch` twice.

alfa-jpn: How about this?

``` ruby
def touch_paranoia_column(with_transaction=false)
  if with_transaction
    with_transaction_returning_status { touch(paranoia_column) }
  else
    touch(paranoia_column)
  end
end
```

or

``` ruby
def touch_paranoia_column(with_transaction=false)
  return with_transaction_returning_status { touch_paranoia_column(false) } if with_transaction
  touch(paranoia_column)
end
```

radar: The first example looks the cleanest to me. Could you please implement that?

The second example has exactly the same bug of calling `touch` twice, so no that is not a good solution.

alfa-jpn:
I changed touch_paranoa_column to the discussed code.

radar:
Well, the CI server seems to be happy with this and the code looks fine to me. Thanks!

radar:
Added to master and rails4.

PR #86, opened 2013-11-14
Separate filter to a method

raszi:
We have a use case where we add filterings on the default scope. We are using that default scope on the public pages but on the admin pages we want to use the unscoped view of the model without the deleted items. Right now we are not able to do that without duplicating the filter in our controller. 

Example after pull request applied:

``` ruby
class Example < ActiveRecord::Base
  default_scope { where(available: true, visible: true).order(:plan_order) }
end

class ExampleController < ApplicationController
  def index
    Example.unscoped.without_deleted
  end
end
```

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -119,7 +119,7 @@ class ActiveRecord::Base
     class_attribute :paranoia_column
 
     self.paranoia_column = options[:column] || :deleted_at
-    default_scope { without_deleted } unless options[:default_scope] == false
+    default_scope { without_deleted }
 
     before_restore {
       self.class.notify_observers(:before_restore, self) if self.class.respond_to?(:notify_observers)
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -14,6 +14,11 @@ module Paranoia
       scoped.tap { |x| x.default_scoped = false }
     end
 
+    def without_deleted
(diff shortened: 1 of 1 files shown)


radar:
Hi @raszi, thanks for the PR.

> Right now we are not able to do that without duplicating the filter in our controller.

What do you mean by this?

raszi:
That in every model in which we want to use the specified filtering we need to use something like the following code:

``` ruby
Model.unscoped.where(Model.paranoia_column => nil))
```

raszi:
I rebased my modifications. Any thoughts on this?

radar:
@raszi In your controller, can't you just call `Example.with_deleted`?

raszi:
We want to have the **not** deleted items not with the deleted.

radar:
I would recommend not overriding the `default_scope` on your model if you want to use paranoia. Only one thing should be defining a default scope.

raszi:
That's not possible in our use-case we need extra filters in our `default_scope`.

Does it really hurt to add this extra scope?

radar:
It adds more code to maintain for me, and I don't see it being a large problem for all people.

raszi:
It uses the same code you wrote. It just exposes the filter to the public.

But okay, it's your repo, your call. We'll use my fork.

PR #88, opened 2013-12-09
Notify observers of paranoid restores

arsduo:
This pull request adds before and after restore hooks to notify ActiveRecord Observers if observers are available (e.g. if `notify_observers` is a method). It's compatible with both the built-in Rails 3 observers (which defines notify_observers through [inheritance](https://github.com/rails/rails/blob/3-2-stable/activerecord/lib/active_record/observer.rb#L91) from [ActiveModel::Observers](https://github.com/rails/rails/blob/4cef581fcf83f161edcd681bbda1461d5f7d5f33/activemodel/lib/active_model/observing.rb#L72)) and Rails 4's [rails-observers](https://github.com/rails/rails-observers/blob/86bfe485dc393f821796f8310cf8bfc1fad6ab3d/lib/rails/observers/active_model/observing.rb#L132). For a description of notifying observers about custom actions, see http://blog.crowdint.com/2013/04/23/fun-with-activerecord-observer.html.

Paranoia and observers -- the jokes write themselves!

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -96,10 +96,10 @@ class ActiveRecord::Base
     default_scope { where(self.quoted_table_name + ".#{paranoia_column} IS NULL") }
 
     before_restore {
-      self.class.notify_observers(:before_restore) if self.class.respond_to?(:notify_observers)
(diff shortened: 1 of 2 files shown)


radar:
Seems pretty reasonable to me. Thanks @arsduo!

PR #89, opened 2013-12-16
JRuby and Rubinius support in specs, Travis

petergoldstein:
1. Updated Gemfile and gemspec to allow JRuby and Rubinius.
2. Added JRuby and Rubinius to .travis.yml

I'd also love to see this pulled over the rails4 branch (as that branch doesn't have any Travis integration at all).  Let me know if it would be helpful for me to submit a separate PR for that branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,3 +2,5 @@ language: ruby
 rvm:
   - 1.9.3
   - 2.0.0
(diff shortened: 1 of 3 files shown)


radar:
Thank you @petergoldstein. I've added this commit to the master and rails4 branch now.

PR #90, opened 2013-12-17
Apply default scope depending on `apply_default_scope` option

DanielVartanov:
Hi,

This option allows to handle a case when one wants deleted associations be still findable even if  they are deleted. Very useful for hiding unneeded records without breaking a logic of other associated models.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,20 @@ class Client < ActiveRecord::Base
 end
 ```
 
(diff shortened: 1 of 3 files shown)


radar:
No.

Override the finder method in the model instead:

https://github.com/spree/spree/blob/2d8bc75d895582e1c1ffab62a9db15800cf54990/core/app/models/spree/variant.rb#L137-L139

PR #91, opened 2013-12-17
Add :recursive option on #restore! to restore associations

remi:
This pull request adds a `recursive` option to the `#restore!` and `.restore` methods so associated records (that were soft deleted) will be restored too when we restore the record they belong to.

``` ruby
  acts_as_paranoid
  has_many :comments, dependent: :destroy
end

class Comment < ActiveRecord::Base
  acts_as_paranoid
  belongs_to :user
end

user = User.create
3.times { user.comments.create }

user.destroy
user.destroyed? # => true
user.comments.count? # => 0

user.restore! :recursive => true
user.destroyed? # => false
user.comments.count # => 3
```

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -19,11 +19,11 @@ module Paranoia
     end
     alias :deleted :only_deleted
 
-    def restore(id)
+    def restore(id, opts = {})
       if id.is_a?(Array)
-        id.map { |one_id| restore(one_id) }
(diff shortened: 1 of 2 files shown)


radar:
Great idea. Thanks @remiprev. CI is happy with it too!

oTangVinhDuong:
paranoia overrides some method 
including ```user.destroyed?```
==> ```user.deleted?```

PR #92, opened 2013-12-19
Add never delete option to rails 3 paranoia

smdern:
The default behavior of for `#destroy` will delete a record if it's called twice. It seems a bit silly for `#destroy`'s behavior to change based on the number of calls, but alas that has become the accepted default.

I've added a `:never_delete` option that will make it so `#destroy` never actually deletes a record from the DB. So `#destroy` will update the paranoia column and that's it.

I'm open for changing the `:never_delete` name to match with the paranoia abstraction.. maybe `:uber_paranoid` or something.

Any :thought_balloon:s?

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -75,6 +75,7 @@ module Paranoia
   # select and exec delete or soft-delete.
   # @param with_transaction [Boolean] exec with ActiveRecord Transactions, when soft-delete.
   def delete_or_soft_delete(with_transaction=false)
+    return touch_paranoia_column(with_transaction) if never_delete
     destroyed? ? destroy! : touch_paranoia_column(with_transaction)
   end
 
@@ -112,7 +113,9 @@ class ActiveRecord::Base
     alias :delete! :delete
     include Paranoia
     class_attribute :paranoia_column
+    class_attribute :never_delete
 
+    self.never_delete = options.fetch(:never_delete) { false }
     self.paranoia_column = options[:column] || :deleted_at
     default_scope { where(self.quoted_table_name + ".#{paranoia_column} IS NULL") }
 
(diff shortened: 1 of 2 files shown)


radar:
@smdern In what situation would the model's .destroy be called a second time?

smdern:
There might be some race condition where two people try to delete the same model.. I can't think of a real valid scenario, but I'M PARANOID! I guess the fact that `#destroy`'s behavior changes offends my sensibilities. I would have though that `#destroy` touches the `paranoia_column` and `#destroy!` actually deletes the record.

brantwedel:
+1 yikes, I agree with @smdern, this is dangerous in a multi-user or multi-threaded environment. The `destroy` methods function can change from soft delete to hard delete depending on a database column whose state might be unknown at the time of invocation; granted, you can wrap every `record.destroy` in a `record.destroy if !record.destroyed?` but that defeats the purpose of overriding `destroy` in the first place.

radar:
I've reverted the change now. You should call `destroy!` if you want to delete a record. Thanks for raising this.

michaltn:
Just fyi this means there is no support for destroy_all ..  to permanently delete records, which is why I proposed the original change to begin with.

brantwedel:
@michaltn it might make sense to add a `destroy_all!` method to bring back that functionality

michaltn:
@onecsguy yeah that might be ok, for destroy_all use case. (although will be more code)

It still wont fix the other problem of has_many relationships that wont work for permanent deletions. (children are never deleted from the database), and fixing that is not trivial.

I think these are the reason why pretty much every other paranoid gem I saw has destroy on a destroyed record delete it from the database. (although I agree that its not beautiful)

PR #95, opened 2013-12-27
Rails 4.1.0.beta1

huoxito:
A couple changes to run on latest rails. Build still has a couple of failures, I believe they are related to the destroy action on dependent records or to the order the tests run, couldnt figure it out yet.

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -8,7 +8,11 @@ module Paranoia
     def paranoid? ; true ; end
 
     def with_deleted
-      all.tap { |x| x.default_scoped = false }
+      if ActiveRecord::VERSION::STRING >= "4.1"
+        unscope where: paranoia_column
+      else
+        all.tap { |x| x.default_scoped = false }
+      end
     end
 
     def only_deleted
@@ -106,7 +110,7 @@ class ActiveRecord::Base
     class_attribute :paranoia_column
 
     self.paranoia_column = options[:column] || :deleted_at
-    default_scope { where(self.quoted_table_name + ".#{paranoia_column} IS NULL") }
+    default_scope { where(paranoia_column => nil) }
 
     before_restore {
       self.class.notify_observers(:before_restore, self) if self.class.respond_to?(:notify_observers)
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -52,8 +52,9 @@ module Paranoia
   end
 
   # As of Rails 4.1.0 +destroy!+ will no longer remove the record from the db
+  # unless you touch the paranoia column before.
   # We need to override it here otherwise children records might be removed
-  # when thet shouldn't
+  # when they shouldn't
   if ActiveRecord::VERSION::STRING >= "4.1"
     def destroy!
       destroyed? ? super : destroy || raise(ActiveRecord::RecordNotDestroyed)
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -51,6 +51,15 @@ module Paranoia
     run_callbacks(:destroy) { touch_paranoia_column(true) }
   end
 
+  # As of Rails 4.1.0 +destroy!+ will no longer remove the record from the db
+  # We need to override it here otherwise children records might be removed
+  # when thet shouldn't
+  if ActiveRecord::VERSION::STRING >= "4.1"
+    def destroy!
+      destroyed? ? super : destroy || raise(ActiveRecord::RecordNotDestroyed)
+    end
+  end
+
   def delete
     return if new_record?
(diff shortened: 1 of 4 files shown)


huoxito:
all green on rails 4.1.0.beta1 :beers:

lib/paranoia.rb
```diff
@@ -8,7 +8,11 @@ module Query
     def paranoid? ; true ; end
 
     def with_deleted
-      all.tap { |x| x.default_scoped = false }
+      if ActiveRecord::VERSION::STRING =~ /4\.1/
```
radar: Please use `>` version comparsion here too.

lib/paranoia.rb
```diff
@@ -47,6 +51,14 @@ def destroy
     run_callbacks(:destroy) { delete_or_soft_delete(true) }
   end
 
+  # As of Rails 4.1.0 +destroy!+ will only permanently destroy a record if it's
```
radar: Please review the changes I did for #92 and edit this.

test/paranoia_test.rb
```diff
@@ -1,13 +1,15 @@
-require 'test/unit'
 require 'active_record'
-require File.expand_path(File.dirname(__FILE__) + "/../lib/paranoia")
-
-DB_FILE = 'tmp/test_db'
 
-FileUtils.mkdir_p File.dirname(DB_FILE)
-FileUtils.rm_f DB_FILE
+test_framework = if ActiveRecord::VERSION::STRING =~ /4\.1/
```
radar: Please use `>` version comparison here.

test/paranoia_test.rb
```diff
@@ -1,13 +1,15 @@
...
+  require 'minitest/autorun'
+  MiniTest::Test
+else
+  require 'test/unit'
+  Test::Unit::TestCase
+end
+require File.expand_path(File.dirname(__FILE__) + "/../lib/paranoia")
 
-ActiveRecord::Base.establish_connection :adapter => 'sqlite3', :database => DB_FILE
+ActiveRecord::Base.establish_connection :adapter => 'sqlite3', database: ':memory:'
```
radar, line 12: :+1: Nice!

test/paranoia_test.rb
```diff
@@ -263,12 +270,16 @@ def test_restore_behavior_for_callbacks
     assert model.instance_variable_get(:@restore_callback_called)
   end
 
-  def test_real_destroy
-    model = ParanoidModel.new
-    model.save
-    model.destroy!
+  # Only relevant to activerecord < 4.0 due to changes in
+  # Associations::HasManyAssociation see https://github.com/rails/rails/commit/5aab0c053832ded70a3a4b58cb97f8f8bba796ba
+  unless ActiveRecord::VERSION::STRING =~ /4\.1/
```
radar: Please use a `>` version comparsion here.

huoxito:
@radar updated. As of the `destroy!` method it will only work as expected if users touch the deleted_at column before calling `destroy!`. Thats the only way I could think of the issue when destroy has_many associations due to changes in https://github.com/rails/rails/commit/5aab0c053832ded70a3a4b58cb97f8f8bba796ba. But maybe there's another way around it?

radar:
@huoxito I don't understand your comment. the `destroy!` method is aliased to ActiveRecord's `destroy` (no bang) method, which will simply attempt to destroy the record... What did you mean with your comment?

huoxito:
Sorry will try to explain again. The issue comes up when you try to _soft_delete_ a product for example (in spree context). It's supposed not to remove the product variants from the database but only touch its deleted_at columns. However if we don't override the `destroy!` implementation here it will remove those records. I think you can confirm that by removing the method and see that one test will fail.

I think the reason is that when destroying the has_many association records rails will now call, see https://github.com/rails/rails/blob/5aab0c053832ded70a3a4b58cb97f8f8bba796ba/activerecord/lib/active_record/associations/has_many_association.rb#L11, the activerecord/persistence `destroy!` method which in turn will call the activerecord/persistence `destroy` and so delete the record from database. hope that makes sense.

radar:
I think you mean this line: https://github.com/rails/rails/blob/5aab0c053832ded70a3a4b58cb97f8f8bba796ba/activerecord/lib/active_record/associations/has_many_association.rb#L110-L111.

In that case, then yes that is very bad. Maybe we should call the method `i_am_really_sure_i_want_to_delete_this!` or something.

huoxito:
@radar just rebased on top of rails4 branch again and added the recently created `really_destroy!` method to the tests. hopefully good to go now

radar:
Added to rails4. Thanks @huoxito.

PR #98, opened 2014-01-06
destroy! now once again calls destroy

radar:
This is because destroy! is called by :dependent => :destroy association definitions (yes, destroy! and not destroy), which means that associated records of a paranoid object would suddenly go missing... making it very hard to restore them!

So instead, it's now called really_destroy! and destroy! is just an alias for destroy once again.

/cc @huoxito

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ Paranoia is a re-implementation of [acts\_as\_paranoid](http://github.com/techno
 
 You would use either plugin / gem if you wished that when you called `destroy` on an Active Record object that it didn't actually destroy it, but just "hid" the record. Paranoia does this by setting a `deleted_at` field to the current time when you `destroy` a record, and hides it by scoping all queries on your model to only include records which do not have a `deleted_at` field.
 
-If you wish to actually destroy an object you may call destroy! on it or simply call destroy twice on the same object.
+If you wish to actually destroy an object you may call `really_destroy!`.
 
 ## Installation & Usage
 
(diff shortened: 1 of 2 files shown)


radar:
@huoxito Do these changes look alright to you, as per the discussion from #95?

huoxito:
Im not sure I understand the alias use but I do like the idea of a third
method to really delete records. Will try to take a look later tonight
thanks mate

Washington L Braga Jr
Developer
Spree Commerce Inc
Em 06/01/2014 03:08, "Ryan Bigg" notifications@github.com escreveu:

huoxito:
Just did some testing and I believe I get it now.

Once you merge it. I could change my PR to use the `really_destroy!` on the tests and we should get a all green build:

``` ruby
  def test_real_destroy
    model = ParanoidModel.new
    model.save
    model.really_destroy!
    refute ParanoidModel.unscoped.exists?(model.id)
  end
```

so it looks pretty good to me. :shipit:

melcher:
Sorry if this should be in a separate ticket. My comment/question is regarding the `#destroy` and `#destroy!` method behavior so it seemed appropriate.

My application depends on `#destroy!` raising an exception if the destroy fails, and i use before_destroy's to hold my business logic of whether something is destroyable.
I want to use paranoia, but reading the documentation was reluctant to have to switch my app from `object.destroy!` over to something like:

``` ruby
object.destroy or raise ActiveRecord::RecordNotDestroyed
```

It seems like this PR would fix the issue for me, but from @radar's comment that "...destroy! is just an alias for destroy once again." it wasn't totally clear.

Anyway, I really like idea behind this pull request, and hope it means folks like me can drop paranoia into our apps, stop our objects from being destroyed, and otherwise keep the behavior of '#destroy' and '#destroy!'.
Thanks, 
graham

Edit: Nope, destroy! is literally an alias for destroy. Any interest (besides myself) in a PR that maintains the rails-ey behavior?

huoxito:
@melcher see the PR mentioned above it adds that very same implementation you described. We can't use destroy! though but `really_destroy!` should do the same job (considering both this and the other pr gets merged)

radar:
Done! Thanks @huoxito.

PR #101, opened 2014-01-15
Tweak, even more clearer

TigerWolf:
More testing and clarification of the behaviour.


radar:
Added! Thanks @TigerWolf

PR #103, opened 2014-01-20
Improved behavior of restore.

alfa-jpn:
Hello.

I'm afraid my expressions may be rude or hard to read, because I'm not so good at English.

When use restore!(:recursive => true), all associations are restored.
Even if model who destroyed before parent is destroyed, restored.
For example,

``` ruby
class ParentModel < ActiveRecord::Base
  has_many :child_models, dependent: :destroy
end

class ChildModel < ActiveRecord::Base
  acts_as_paranoid
  belongs_to :parent_model
end
```

When call restore!, The changes as follows.

``` ruby
parent  = ParentModel.create
first      = parent.child_models.create
second = parent.child_models.create

first.destroy
# first => destroyed

parent.destroy
# parent => destroyed, second => destroyed

parent.restore!(:recursive => true)
# parent => restored, second => restored,  and ... first => restored

```

This pr improved behavior of restore.
If child model has a dependent_delete column(*modifiable), dependent_delete has a info of 'deleted by dependence'. restore method see it.

``` sh
rails g add_dependent_delete_to_child_models dependent_delete:boolean
```

dependent_delete column is modifiable.

``` ruby
acts_as_paranoid dependent_column: dependent_destroyed
```

When call restore!, The changes as follows.

``` ruby
parent  = ParentModel.new
first      = parent.child_models.create
second = parent.child_models.create

first.destroy
# first => destroyed

parent.destroy
# parent => destroyed, second => destroyed

parent.restore!(:recursive => true)
# parent => restored, second => restored

```

In addition, do i have to be send pullrequest for branch of rails4.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ pkg/*
 tmp
 .rvmrc
 Gemfile.lock
+.project
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -77,32 +77,48 @@ module Paranoia
   # set dependent delete flg to associations.
   # @note This method will be called when run before_destroy event.
   def set_dependent_associations
-    each_paranoid_associations do |association|
-      if association.klass.column_names.include?(paranoia_dependent_column.to_s)
-        association.where(deleted_at: nil).update_all(paranoia_dependent_column => true)
+    each_paranoid_associations do |resource, resource_class, collection|
+      next unless resource_class.column_names.include?(paranoia_dependent_column.to_s)
+
+      if collection
+        resource.where(deleted_at: nil).update_all(paranoia_dependent_column => true)
+      else
+        resource.update_column(paranoia_dependent_column, true)
       end
     end
   end
 
   # exec block with each paranoid association.
-  # @param &block [Proc{|association| .. }] exec block.
+  # @param &block [Proc{|resource, resource_class, collection_flg| .. }] exec block.
   def each_paranoid_associations(&block)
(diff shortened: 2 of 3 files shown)


alfa-jpn:
I found a new bug. 

raise error, when recover associate by belongs or has_one.

I will add  the commit of this bugfix.

alfa-jpn:
Fixed the next bug
- Can't destroy dependent model that have one-to-one relationship.(belongs_to and has_one).
  
  **paranoia v2.0.2 has this bug too.**

``` ruby
Class Parent
  belongs_to :child, :dependent => :destroy
end

parent = Parent.create
child   = Child.create(parent_id: parent.id)

parent.reload.destroy  # raise error.
```

radar:
Hi @alfa-jpn. I understand what bug you're seeing here. I think it is normal behaviour for the child object to be restored if you call restore on the parent immediately after calling destroy on the child. I don't see that as a bug. I see that as intentional behaviour.

I think adding this extra column to paranoia to fix this bug is introducing too much code to a library that should be staying fairly light. If there's a super-serious bug that's breaking the library, then I would gladly accept any amount of code to fix it. However, this seems to be an extreme case and only reproducible under specific circumstances. Therefore I do not think it belongs in paranoia.

Thank you for submitting a PR anyway.

rx-8:
This account is other accounts of alfa-jpn.

Extra column is optional.
Even if model doesn't have it, paranoia don't raise error.

but does this PR be rejected?

PR #104, opened 2014-01-25
Fix destroy return value.

kouno:
Rails normally returns the model on successful destroy and false if any
callback fails.


radar:
Thanks @kouno for the PR **with tests**! I've added this to master and rails4 now.

PR #105, opened 2014-02-01
Update README

davidpiegza:
Just a small README update, removing the text `or simply call destroy twice on the same object`.


radar:
Thanks @davidpiegza :)

PR #106, opened 2014-02-04
fixed restore recursive association for rails 3

bublik:


bublik:
Do you think my fix is not correct ?

radar:
I don't know why this was closed... I don't remember closing it.

Could you please write a regression test for this as well?

PR #107, opened 2014-02-06
Recovery window

howardr:
Added a recovery window to restore associated records that were deleted within a certain timespan of a parent object deletion.

This is mimicking similar behavior in [acts_as_paranoid](https://github.com/goncalossilva/acts_as_paranoid) that we needed.

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -47,7 +47,10 @@ module Paranoia
   end
 
   def destroy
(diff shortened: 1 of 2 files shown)


radar:
I think this is outside the scope of the basic functionality that paranoia is offering. Thank you for your contribution, but I think because this is such a limited use-case for people, it's probably best left in your own fork.

Sorry!

PR #113, opened 2014-02-13
Add index to deleted_at column in README

rvanlieshout:
Shouldn't we add an index on deleted_at?


codeodor:
:thumbsup:

radar:
I didn't know about the `:index` syntax. Cool.

PR #118, opened 2014-02-27
Fixes an exception when a has_one association is recursively restored.

stupergenius:
The original implementation would raise a `method_missing` exception when a has_one association was recursively restored. The pull request first checks to see if the association is a collection before looping through the deleted items.

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -103,10 +103,14 @@ module Paranoia
     end
 
     destroyed_associations.each do |association|
-      association = send(association.name)
+      association_data = send(association.name)
 
-      if association.paranoid?
-        association.only_deleted.each { |record| record.restore(:recursive => true) }
+      if association_data.paranoid?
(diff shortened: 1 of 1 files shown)


radar:
Hi @stupergenius, could you please write a regression test for this?

pobocks:
Am I correct in thinking that the case of a nullable single collection should also be handled?  i.e., if association_data is nil, currently restore fails.

My additional fix, off the top of my head, probably wrong:

``` ruby
...
association_data = send(association.name)
unless association_data.nil?
  if association_data.paranoid?
    if association.collection?
      association_data.only_deleted.each { |record| record.restore(:recursive => true) }
...
```

pobocks:
The nil check seems to fix the remaining issue.  I've made a new pull request with the combined fix.

https://github.com/radar/paranoia/pull/119

radar:
Fixed by #119.

PR #119, opened 2014-03-07
Regression tests and full fix for #118

pobocks:
- Based on partial fix by @stupergenius
- Handles nil case for hasOne associations


radar:
Added to rails3 and rails4. Thanks @stupergenius and @pobocks.

pobocks:
Glad to be of help!

Issue #114, opened 2014-02-13
validates_uniqueness_of does not use "where deleted_at is null"


codeodor:
It's trivial in Rails 4 for the client code to fix by adding `:conditions` but that's not an option in Rails 3.

Is there any interest in a fix for this, either by adding a new method (maybe append `_without_deleted` to the name, or fixing it to work to be consistent with the rest of the queries it modifies?

radar:
@codeodor Can you show me a situation with this failing?

codeodor:
Sure, I added tests to each branch that show it failing.

`Employer` has a new column `name` on which we use `validates_uniqueness_of` in the model.

The test creates a record, destroys it, and creates a new instance with the same name, and asserts whether the record is valid. At present, it is not.

kidlab:
I encountered this problem too.
Here is my models:

```
class User < ActiveRecord::Base
  acts_as_paranoid

  has_many :favorites, dependent: :destroy
end

class Book < ActiveRecord::Base
  acts_as_paranoid

  has_many :favorites, dependent: :destroy
end

class Favorite < ActiveRecord::Base
  acts_as_paranoid

  belongs_to :user
  belongs_to :book

  validates :user, :book, presence: true
  validates_uniqueness_of :book_id, scope: :user_id
end
```

This code below will add book to user's favorite then remove it by soft-deleting. After that I cannot add book to favorite any more, because `validates_uniqueness_of :book_id, scope: :user_id` does not use paranoid scope.

```
user = User.first
book = Book.first

# Add my favorite book
f = Favorite.create(user: user, book: book)
# => #<Favorite id: 1, user_id: 1, book_id: 1, deleted_at: nil, created_at: "2014-03-15 17:53:50", updated_at: "2014-03-15 17:53:50">

# Remove from favorite
f.destroy

# Re-add book to favorite
f = Favorite.new(user: user, book: book)
f.valid?
# => false

f.errors
# => #<ActiveModel::Errors:0x007f2a8922acf8 @base=#<Favorite id: nil, user_id: 1, book_id: 1, deleted_at: nil, created_at: nil, updated_at: nil>, @messages={:book_id=>["has already been taken"]}>
```

smostovoy:
+1. https://github.com/goncalossilva/acts_as_paranoid has validates_uniqueness_of_without_deleted. IMHO it should be default. @radar do you plan anything?

codeodor:
[...]

PR #121, opened 2014-03-12
Fix Paranoia to work with validates_uniquness_of (rails4 branch)

codeodor:
For the rails4 branch, in pursuit of #114

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -237,3 +237,16 @@ class ActiveRecord::Base
 end
 
 require 'paranoia/rspec' if defined? RSpec
+
+module ActiveRecord
+  module Validations
+    class UniquenessValidator < ActiveModel::EachValidator
+      protected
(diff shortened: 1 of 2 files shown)


codeodor:
@radar just wondering if you had a chance to look at this PR and the one in #121 too. I got bit by this again, so it reminded me of these. :smile:

ethangunderson:
Any update on this one? I never learn, so this keeps biting us.

radar:
If there was an update you can be sure it would be posted here. There is no secret place where things about this gem get discussed. It's all done here out in the open.

ethangunderson:
I apologize, that was a poorly worded comment on my part. Let me try again. I would like to see this merged, is there anything I can do to help it along?

codeodor:
This one should be ready now too.

codeodor:
This will also fix #175

PR #122, opened 2014-03-12
Fix Paranoia to work with validates_uniquness_of (rails3 branch)

codeodor:
For the rails3 branch, in pursuit of #114

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -172,4 +172,18 @@ class ActiveRecord::Base
   end
 end
 
+
 require 'paranoia/rspec' if defined? RSpec
+
+module ActiveRecord
+  module Validations
+    class UniquenessValidator < ActiveModel::EachValidator
+      protected
+      def build_relation_with_paranoia(klass, table, attribute, value)
+        relation = build_relation_without_paranoia(klass, table, attribute, value)
+        relation.and(klass.quoted_table_name + ".#{klass.paranoia_column} IS NULL")
+      end
+      alias_method_chain :build_relation, :paranoia
+    end
+  end
+end
--- a/paranoia.gemspec
+++ b/paranoia.gemspec
@@ -18,6 +18,7 @@ Gem::Specification.new do |s|
 
   s.add_development_dependency 'bundler', '>= 1.0.0'
   s.add_development_dependency 'rake'
+  s.add_development_dependency 'test-unit'
 
   s.files = `git ls-files`.split("\n")
   s.executables = `git ls-files`.split("\n").map { |f| f =~ /^bin\/(.*)/ ? $1 : nil }.compact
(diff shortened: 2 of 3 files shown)


radar:
@codeodor The build on this PR passes. Does that mean that you've fixed it?

codeodor:
It was #121 that was failing, I guess. Locally it passes all the tests on Rails 4.0.2 and Ruby 2.0, but [the build shows failures](https://travis-ci.org/radar/paranoia/jobs/20878945) due to `assert_nothing_raised` not being there in Rails 4.1, and something wrong with `test_delete_behavior_for_plain_models_callbacks`. 

For the failing test, I have no idea why it would pass locally and on the 4.1 tests. For the error, I think that test needs to be rewritten (but has nothing to do with this PR).

And actually, now that I look at [the most recent build](https://travis-ci.org/radar/paranoia/builds/21149688) it appears to have the same two issues, so I'd say both are ready to go, with the two tests being a separate issue.

codeodor:
Oh, and as far as "fixed it" then yes, I did fix the issue raised in #114 on both the Rails 3 branch (this PR) and the Rails 4 branch (#121).

At first I thought by "fixed it" you might have been referring to the build failures on the Rails 4 branch that I [mentioned in the original issue](https://github.com/radar/paranoia/issues/114#issuecomment-37758389).

joshuapinter:
Any status update as to when this might get merged in? 

Thanks!

radar:
screen shot 2015-06-18 at 7 31 24 am

PR will need to be rebased on the rails3 branch before it can be merged in.

codeodor:
I will rebase it when I get a chance.

codeodor:
Ok, all done.

codeodor:
@radar I also rebased #121 for the Rails 4 branch.

codeodor:
This will also fix #175

radar:
Travis is happy. I am happy. Let's merge it.

codeodor:
:+1:  Thank you!

joshuapinter:
Not sure of the details yet, but after using this merged version my uniqueness validation stopped working. It allows me to add multiple records with the same value. 

I know as part of this PR there's a test to ensure that `validates_uniqueness_of` ignores rows where `deleted_at` is not NULL, but is there a complimentary test to ensure that `validates_uniqueness_of` still works as it should?

joshuapinter:
So, change: 

``` ruby
def test_validates_uniqueness_only_checks_non_deleted_records
  a = Employer.create!(name: "A")
  a.destroy
  b = Employer.new(name: "A")
  assert b.valid?
end
```

To this: 

``` ruby
def test_validates_uniqueness_only_checks_non_deleted_records
  a = Employer.create!(name: "A")
  b = Employer.new(name: "A")

  assert_not b.valid?

  a.destroy
  assert b.valid?
end
```

codeodor:
Created #246 to deal with that. Thanks @joshuapinter

joshuapinter:
:+1: Beauty. Thanks @codeodor.

noahmatisoff:
Had same problem, updated gem, fixed it.  Thanks.

ioev:
Looks like this patch will raise an exception for any model that does not "acts_as_paranoid", as they will not have a paranoia_column defined.  I fixed this locally with:

`klass.paranoid? ? relation.and(klass.arel_table[klass.paranoia_column].eq(nil)) : relation`

codeodor:
@ioev is that what #286 is for?

Edit: oh, probably not. Looks like they just revert it to its old broken behavior instead of the new broken behavior.

ioev:
I'm not sure, but I don't think so.  I'm assuming that the rails 4 handles this in a different way, and so this bit of code became unnecessary?

Sorry, I should have been more specific and mentioned that I'm looking at code from the Rails 3 branch, as that is what the project I'm working on is.

codeodor:
I'd encourage a PR. I dunno if they plan to keep up any compatibility with Rails 3, but it definitely was a big oversight on my part in the original PR.

ioev:
Thanks, I'll see if I can find the time to put one together.

Issue #120, opened 2014-03-12
About the  has_one association problem


kai209209:
![catch](https://f.cloud.github.com/assets/5562823/2395250/94ac5f14-a9c6-11e3-8067-5199850fc0f8.png)

My Gemfile file use
gem 'paranoia', :github => 'radar/paranoia', :branch => 'rails4' 

 User model
  has_one :organization, dependent: :destroy
  has_many :activities, dependent: :destroy

Organization model
  belongs_to :user

Activity model
  belongs_to :user

The user has_one organization , and user has_many activities.
Once I restore the user record, execute  "User.restore(10, :recursive => true)", all activities are ok, but the organization still missing. 
Look at the picture, the query :  
SELECT `organizations`.\* FROM `organizations` WHERE `organizations`.`deleted_at` IS NULL AND `organizations`.`user_id` = 10 ORDER BY `organizations`.`id` ASC LIMIT 1  

Why it  found  the record :  'deleted_at' IS NULL
The correct should be "'deleted_at' IS NOT NULL " ,right?
So, is it a bug?

kai209209:
my rails version is  '4.0.2'

radar:
@kai209209 I don't think that it'll fix this particular issue, but you should really upgrade to 4.0.3, since that's the latest release and contains security problems.

radar:
I believe #123 fixes this problem.

PR #123, opened 2014-03-13
hasOne association could not restore

JeskTop:
When I find @pobocks fix the problem https://github.com/radar/paranoia/pull/119 , and I use the new paranoia and find another bug. 
The problem can look at my first commit what is a test and the problem like https://github.com/radar/paranoia/issues/120 .

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -114,6 +114,10 @@ module Paranoia
           end
         end
       end
+
+      if association_data.nil? && association.macro.to_s == "has_one"
+        Object.const_get(association.name.to_s.camelize).only_deleted.where("#{self.class.name.to_s.underscore}_id", self.id).first.try(:restore, recursive: true)
+      end
     end
   end
 end
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -123,7 +123,9 @@ module Paranoia
       end
 
       if association_data.nil? && association.macro.to_s == "has_one"
-        Object.const_get(association.name.to_s.camelize).only_deleted.where("#{self.class.name.to_s.underscore}_id", self.id).first.try(:restore, recursive: true)
+        association_class_name = association.options[:class_name].present? ? association.options[:class_name] : association.name.to_s.camelize
+        association_foreign_key = association.options[:foreign_key].present? ? association.options[:foreign_key] : "#{self.class.name.to_s.underscore}_id"
+        Object.const_get(association_class_name).only_deleted.where(association_foreign_key, self.id).first.try(:restore, recursive: true)
       end
     end
   end
--- a/test/paranoia_test.rb
+++ b/test/paranoia_test.rb
@@ -387,6 +387,47 @@ class ParanoiaTest < test_framework
     assert ParanoidModelWithBelong.with_deleted.reload.count != 0, "There should be a record"
   end
 
+  def test_new_restore_with_has_one_association
+    # setup and destroy test objects
+    hasOne = ParanoidModelWithHasOne.create
+    belongsTo = ParanoidModelWithBelong.create
+    hasOne.paranoid_model_with_belong = belongsTo
+    hasOne.save!
+
+    hasOne.destroy
+    assert_equal false, hasOne.deleted_at.nil?
+    assert_equal false, belongsTo.deleted_at.nil?
+
+    # Does it restore has_one associations?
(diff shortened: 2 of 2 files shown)


lib/paranoia.rb
```diff
@@ -114,6 +114,10 @@ def restore_associated_records
           end
         end
       end
+
+      if association_data.nil? && association.macro.to_s == "has_one"
+        Object.const_get(association.name.to_s.camelize).only_deleted.where("#{self.class.name.to_s.underscore}_id", self.id).first.try(:restore, recursive: true)
```
radar: Um, wow.

Could you please write some documentation about this? There's a lot of meta programming that goes on within this one line!

JeskTop:
My commits fix the problem is the `has_one` counld not restore. 
I submitted before the test is passed, because the destroy object did not find from the database and it just a catch or not. Code: https://github.com/radar/paranoia/blob/rails4/test/paranoia_test.rb#L382

JeskTop:
How can I find the has_one object?

The difference between `has_one` and `has_many` like this:

``` ruby
# User has_many Message
@user.messages.where(title: "abc")   #This is can work!
```

``` ruby
# User has_one Role
@user.role.where(name: "abc")  #This is can not work!
```

So if I want to find the `role` must use:

``` ruby
Role.where(name: "abc", user_id: @user.id)
```

It just like:

``` ruby
Object.const_get("role").where(name: "abc", user_id: @user.id)
```

So this one line:

``` ruby
Object.const_get(association.name.to_s.camelize).only_deleted.where("#{self.class.name.to_s.underscore}_id", self.id).first.try(:restore, recursive: true)
```

This mean is use `association.name.to_s.camelize` find the `has_one` model name and use `self.class.name.to_s.underscore` to setting this class `id` (`user_id`).

And then user `User has_one Role` it will work look like:

``` ruby
Object.const_get("role").only_deleted.where("user_id", self.id).first.try(:restore, recursive: true)
```

Or

``` ruby
Role.only_deleted.where("user_id", self.id).first.try(:restore, recursive: true)
```

lib/paranoia.rb
JeskTop: Hi @radar , my English is not very good, write to the following several documentations, do not know whether or not clear?

pobocks:
Is this guaranteed to be the fkey name in question?

``` ruby
{self.class.name.to_s.underscore}_id
```

It seems like there'd need to be a check for key columns specified in the options.

JeskTop:
@pobocks Oh. Yes! I must think about how to use the `foreign_key` to find the model.

JeskTop:
Hey, @pobocks. I pushed the new commit and checked the specified `foreign_key` and `class_name`.

adityashedge:
I'm having an issue with restoring dependent has_one relation recursively in rails 4.1. Can you check if this merge request fixes it and if it does please merge it ASAP. Thank  you in advance.

Exact Problem:
User has_one UserInfo and UserInfo belongs_to User.
Unable to restore UserInfo when restoring user and all dependent records recursively.
UserInfo is soft deleted but not restored when restoring User recursively.

JeskTop:
@adityashedge 
Thanks. I think the commit can fix all of your problems.
You can see my test. If `User has_one UserInfo and UserInfo belongs_to User`, It will restore `UserInfo` when restoring user and all dependent records recursively.
And if you find the another problem, you can tell me.

willsu:
+1 @JeskTop This fixed my issue with `restore(recursive: true)` on a has_one association.

@radar Any chance of getting this pull request accepted? I'm pointing my Gemfile at the @JeskTop fork and `setting_restore` branch for the time being.

radar:
The PR no longer applies cleanly to the targeted branch. If this PR got an update to fix that, then I would gladly pull this in.

JeskTop:
What do you mean clean? I alreay update the `rails4` branch and merge for my `setting_restore` branch. If @radar has another problems, you can tell me and I will fix that. Thanks.

ccaratori:
Hey guys is this going to be merged and released any time soon?

radar:
@JeskTop I still can't automatically merge this PR, and it has failing tests.

screen shot 2014-09-15 at 11 18 38 am

Once those failing tests and merge conflicts are fixed, then I can look at this again.

JeskTop:
@radar Thanks. I already fix all conflicts. And push the commit.

radar:
Thank you very much @JeskTop.

PR #124, opened 2014-03-13
Test case for broken counter caches

howardr:
Counter caches are not updating correctly after a child association is restored. If I try to fix this in the model by using an `after_restore` hook, then a different test case breaks.

### DIFF
--- a/test/paranoia_test.rb
+++ b/test/paranoia_test.rb
@@ -8,7 +8,7 @@ FileUtils.mkdir_p File.dirname(DB_FILE)
 FileUtils.rm_f DB_FILE
 
 ActiveRecord::Base.establish_connection :adapter => 'sqlite3', :database => DB_FILE
-ActiveRecord::Base.connection.execute 'CREATE TABLE parent_models (id INTEGER NOT NULL PRIMARY KEY, deleted_at DATETIME)'
+ActiveRecord::Base.connection.execute 'CREATE TABLE parent_models (id INTEGER NOT NULL PRIMARY KEY, deleted_at DATETIME, related_models_count INTEGER)'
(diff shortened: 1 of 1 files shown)


test/paranoia_test.rb
```diff
@@ -400,6 +402,38 @@ def test_observers_not_notified_if_not_supported
...
+
+    # destroy child and check counter cache on parent
+    child.destroy
+    parent.reload
+    assert_equal 0, parent.related_models_count
+
+    # restore child and check
+    child.restore!
+    parent.reload
+    assert_equal 1, parent.related_models_count
```
howardr, line 435: Specifically this one is broken right now. 

```
<1> expected but was
<0>.
```

radar:
@howardr Thanks for the test to reproduce the issue. Have you had any luck with fixing it yet?

bluej100:
@radar : What would you think of triggering after_create as well as after_restore when restoring? I feel like it is generally equivalent to a creation. I'm seeing the same counters issue with https://github.com/magnusvk/counter_culture .

radar:
No, I think that's a bad approach. The record is not created. It is only restored. Please investigate an alternativez

bluej100:
I disagree. Restoring is a type of creation, as it adds a new record
matching a previously deleted record to the default scope. It is the
opposite of destroy. It introduces a new record which matches a deleted
record. In my opinion, the fact that this gem keeps rows of deleted records
in the same table should be thought of as an implementation detail: they
could just as well be kept in parallel archive tables.

If this gem is to interoperate with others that don't have a concept of
soft deletion, I think triggering a create event is the best solution.

BenMorganIO:
I think @excid3 just fixed this. Let me know if I'm at error :)

PR #125, opened 2014-03-17
Explicitly load AR unless already loaded

andrewslotin:
This fixes `uninitialized constant Paranoia::ActiveRecord` when using `paranoia` in a non-Rails application.


radar:
Thanks @andrewslotin, I've added this to rails3 and rails4 now.

PR #126, opened 2014-03-18
Add rspec matcher: act_as_paranoid

mattconnolly:
This provides as RSpec matcher that can be used like this in models:

``` ruby
describe Model
  it { should act_as_paranoid }
end
```


radar:
Thanks @mattconnolly. I've added this to rails3 and rails4.

mattconnolly:
Thank you!

PR #128, opened 2014-04-11
Adds #really_destroyed? method to work as default rails #destroyed?

yratanov:
Paranoia gem reimplements `#destroyed?` method to check only `paranoia_column` presence. But sometimes we need to make sure record is going through real destroy process. 
For example if we have attached file and we want to destroy it only on real destroy:

``` ruby
after_destroy :remove_attachments, if: :really_destroyed?
```

New method `#really_destroyed?` added to check if model destroyed completely.


radar:
Thanks!

PR #132, opened 2014-05-10
Delete all doesn't soft delete

RobinClowers:
I'm not sure if this is expected behavior, but it certainly caught me off guard. I started to look into how to fix it, but I'm not that familiar with AR.

### DIFF
--- a/test/paranoia_test.rb
+++ b/test/paranoia_test.rb
@@ -128,6 +128,21 @@ class ParanoiaTest < test_framework
     assert_equal 1, model.class.unscoped.count
   end
 
+  def test_delete_all_behavior_for_paranoid_models
+    model = ParanoidModel.new
+    model2 = ParanoidModel.new
+    assert_equal 0, model.class.count
+    model.save!
+    model2.save!
+    assert_equal 2, model.class.count
+    ParanoidModel.delete_all
+
+    assert_equal false, model.reload.deleted_at.nil?
+
+    assert_equal 0, model.class.count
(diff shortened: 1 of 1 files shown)


test/paranoia_test.rb
```diff
@@ -99,7 +99,7 @@ def test_delete_behavior_for_plain_models_callbacks
     assert_equal nil, model.instance_variable_get(:@validate_called)
     assert_equal nil, model.instance_variable_get(:@destroy_callback_called)
     assert_equal nil, model.instance_variable_get(:@after_destroy_callback_called)
-    assert_equal nil, model.instance_variable_get(:@after_commit_callback_called)
+    # assert_equal nil, model.instance_variable_get(:@after_commit_callback_called)
```
radar: I would rather this line not be commented out... Either remove it (with an explanation of why it was removed) or keep it.

RobinClowers: I commented it out because it was failing before I started. It seemed like it should pass, so I didn't want to remove it.

radar: Ok. I would rather it be left failing if it was failing before hand. It's likely that a change in Rails made this happen, and so if that's the case then a failing test indicating that Something's Not Right(tm) is better than wilful ignorance.

RobinClowers: Good point, I just rebased that commit out.

jhawthorn:
This is expected behaviour, or at least it's behaviour which would be problematic to change.

We could do a better job of documenting which methods works as soft-deletes. I've opened #304

PR #134, opened 2014-06-05
Add note about different behaviors of `recover` vs `restore`.

dhamidi:
The `acts_as_paranoid` gem (v0.4.1) uses assignment followed by a call
to `#save` for restoring a record:

```
# in #recover
run_callbacks :recover do
  recover_dependent_associations(options[:recovery_window], options) if options[:recursive]

  self.paranoid_value = nil
  self.save
end
```

This runs any `update` callbacks as well as `recover` callbacks.

Paranoia uses `#update_column` instead of `#save`, so no `update`
callbacks are run when restoring a record.

With this commit the difference in behavior is described in the section
"Acts As Paranoid Migration" of the README.


radar:
Thank you @dhamidi. I've added this commit to rails4 now.

PR #135, opened 2014-06-08
Allow destroying a non-persisted Paranoid object

lowjoel:
Currently destruction is made via touching the paranoid column. This works unless the object being destroyed has not yet been persisted, such as when replacing the other end of a has_one relation. There is no need to mark it destroyed in that scenario because it will not be stored in database.

### DIFF
--- a/test/paranoia_test.rb
+++ b/test/paranoia_test.rb
@@ -203,10 +203,10 @@ class ParanoiaTest < test_framework
   end
 
   def test_destroy_behaviour_for_unpersisted_models
-		model = ParanoidModel.new
-		model.destroy
+    model = ParanoidModel.new
+    model.destroy
 
-		# No assertion because we just want to ensure that ActiveRecord does not complain.
+    # No assertion because we just want to ensure that ActiveRecord does not complain.
   end
 
   # Regression test for #24
--- a/test/paranoia_test.rb
+++ b/test/paranoia_test.rb
@@ -202,6 +202,13 @@ class ParanoiaTest < test_framework
     assert_equal 1, model.class.unscoped.count
   end
 
+  def test_destroy_behaviour_for_unpersisted_models
+		model = ParanoidModel.new
+		model.destroy
+
+		# No assertion because we just want to ensure that ActiveRecord does not complain.
+  end
+
   # Regression test for #24
   def test_chaining_for_paranoid_models
     scope = FeaturefulModel.where(:name => "foo").only_deleted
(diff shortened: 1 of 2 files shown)


radar:
Hi @lowjoel, thank you for the patch. Could you please write a regression test for this too?

lowjoel:
ideally we should test for `destroyed?` after `destroy`; but unless we do more major changes this would have to do...

test/paranoia_test.rb
```diff
@@ -171,6 +171,13 @@ def test_destroy_behavior_for_featureful_paranoid_models
     assert_equal 1, model.class.unscoped.count
   end
 
+  def test_destroy_behaviour_for_unpersisted_models
+    model = ParanoidModel.new
+    model.destroy
+
+    # No assertion because we just want to ensure that ActiveRecord does not complain.
```
radar, line 209: Why would AR complain in this context?

lowjoel, line 209: http://api.rubyonrails.org/v4.1.1/classes/ActiveRecord/Persistence.html#method-i-touch

Point 3:

Note that touch must be used on a persisted object, or else an ActiveRecordError will be thrown. For example:

``` ruby
ball = Ball.new
ball.touch(:updated_at)   # => raises ActiveRecordError
```

ragaskar:
Pulling lowjoel's fork fixed the remaining failing tests on our build after going to Rails 4.1 and paranoia head. Looks like we had some bad interactions with some FactoryGirl factories and head of rails4 paranoia.

ragaskar:
Looks like this branch no longer applies. I used the following monkey patch against rails4 head instead:

```
require 'paranoia/version'

unless Paranoia::VERSION == '2.0.2'
  # If you see this message, please test removing this file
  # If it's still required, please bump up the version above
  Rails.logger.warn "Please remove me, Paranoia version has changed"
end

require 'paranoia'
module Paranoia
  private
  def touch_paranoia_column(with_transaction=false)
    # Sometimes paranoid objects are created by has_one relationships; they are
    # never persisted to the database, so touching the column would raise an
    # exception from ActiveRecord.
    return unless persisted? # <--- the key monkey patch bit.

    # This method is (potentially) called from really_destroy
    # The object the method is being called on may be frozen
    # Let's not touch it if it's frozen.
    unless self.frozen?
      if with_transaction
        with_transaction_returning_status { touch(paranoia_column) }
      else
        touch(paranoia_column)
      end
    end
  end
end
```

radar:
@ragaskar Could you please submit a new PR with a patch (that has tests) that fixes this problem?

lowjoel:
@radar I've already put the test and the rationale behind it. I can't think of any better way of testing for it.

ragaskar:
Yep, seems like easiest thing is for the lowjoel branch to rebase against HEAD, should be a simple merge.

lowjoel:
Just rebased. Let's see.

jhawthorn:
Thanks :tada:. This should be fixed in `2.1.0.pre`

PR #136, opened 2014-06-13
Update paranoia_test.rb

mjording:
Deprecated minitest method assert_nothing_raised was blowing up on travis. 

Removed deprecated assert_nothing_raised from test.


radar:
Thank you very much @mjording :)

radar:
I've added this to the rails3 branch as well.

PR #138, opened 2014-06-20
Support an additional boolean column to enable unique key enforcement

n3bulous:
Provide support for handling a boolean column on tables with deleted_at so unique keys can be enforced (null values in a unique key do not allow enforcement of the unique key)

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,8 +180,7 @@ You can replace the older `acts_as_paranoid` methods as follows:
 
 ## Support for Unique Keys with Null Values
 
-With most databases, a unique key will not be enforced when it contains a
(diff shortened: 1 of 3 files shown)


radar:
Looks good to me. Thanks @n3bulous :)

radar:
@n3bulous Would you like to create a PR for the rails3 branch as well?

n3bulous:
I can do that after vacation :)

radar:
Excellent, thank you :)

n3bulous:
I also need to update the documentation a bit.  The index example should list :is_deleted first in the composite key.

PR #139, opened 2014-06-23
Allow rails 3.1

tamird:
So the tests on rails3 don't pass at all (including on 3.2), but the same tests fail on 3.1 and 3.2.

Unfortunately, the latest version of this gem available for 3.1 suffers from https://github.com/radar/paranoia/issue/81. This patch will make fix that issue for rails 3.1 apps.

### DIFF
--- a/paranoia.gemspec
+++ b/paranoia.gemspec
@@ -14,7 +14,7 @@ Gem::Specification.new do |s|
   s.required_rubygems_version = '>= 1.3.6'
   s.rubyforge_project = 'paranoia'
 
-  s.add_dependency 'activerecord', '~> 3.2'
+  s.add_dependency 'activerecord', '~> 3.1'
 
   s.add_development_dependency 'bundler', '>= 1.0.0'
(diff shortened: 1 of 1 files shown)


paranoia.gemspec
```diff
@@ -14,7 +14,7 @@ Gem::Specification.new do |s|
   s.required_rubygems_version = '>= 1.3.6'
   s.rubyforge_project = 'paranoia'
 
-  s.add_dependency 'activerecord', '~> 3.2'
+  s.add_dependency 'activerecord', '>= 3.1'
```
radar: Please make this `~> 3.1`, by using `>=` you're actually allowing any future version of Rails, including a Rails 4, 5 or 6 version.

tamird: done, thanks!

radar:
Thanks @tamird :) I've added this to the rails3 branch.

tamird:
are you sure? don't see it

tamird:
thanks! would you mind cutting a release with this change?

PR #140, opened 2014-06-24
Add optional sentinel_value

smdern:
MySQL ignores null columns when it comes to resolving unique index constraints. This means that all of our unique constraints that involve nullable columns are problematic. This will allow the `deleted_at` value to be set globally or per model.

``` ruby
#pick some epoch
acts_as_paranoid sentinel_value: DateTime.new(0)
```

or globally in a rails initializer

`Paranoia.default_sentinel_value = DateTime.new(0)`

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,44 +182,6 @@ You can replace the older `acts_as_paranoid` methods as follows:
 The `recover` method in `acts_as_paranoid` runs `update` callbacks.  Paranoia's
 `restore` method does not do this.
 
-## Support for Unique Keys with Null Values
-
-With most databases, a unique key containing a null value will not be enforced because the null value is unique per row.
-
-``` ruby
-class AddDeletedAtToClients < ActiveRecord::Migration
-  def change
-    add_column :clients, :deleted_at, :datetime
-    add_index :clients, [:username, :deleted_at], unique: true
-  end
-end
-```
-
-Given the migration above, you could have multiple users with username bob given the following inserts: ('bob', null), ('bob', null), ('bob', null). We can agree this is not the expected behavior.
-
-To avoid this problem, we could use a flag column instead of a datetime, but the datetime value has intrinsic usefulness.  Instead, we can add a second column for the unique key that always has a value, in this case 0 or 1:
-
-``` ruby
-class AddDeletedAtToClients < ActiveRecord::Migration
-  def change
-    add_column :clients, :deleted_at, :datetime
-    add_column :clients, :is_deleted, :boolean, null: false, default: 0
-    add_index :clients, [:username, :is_deleted], unique: true
-  end
-end
-```
-
-Support this new column by updating your model as such:
-
-``` ruby
-class Client < ActiveRecord::Base
-  acts_as_paranoid :flag_column => :is_deleted
-  ...
-end
-```
-
-If you create an index on the flag column, and you want paranoia to use that index instead of deleted_at, you can add `:index_column => :is_deleted` to the acts_as_paranoid definition.
-
 ## License
 
 This gem is released under the MIT license.
--- a/README.md
+++ b/README.md
@@ -182,6 +182,27 @@ You can replace the older `acts_as_paranoid` methods as follows:
 The `recover` method in `acts_as_paranoid` runs `update` callbacks.  Paranoia's
 `restore` method does not do this.
 
+## Support for Unique Keys with Null Values
+
+Most databases ignore null columns when it comes to resolving unique index
+constraints.  This means unique constraints that involve nullable columns may be
+problematic. Instead of using `NULL` to represent a not-deleted row, you can pick
+a value that you want paranoia to mean not deleted. Note that you can/should
+now apply a `NOT NULL` constraint to your `deleted_at` column.
+
+Per model:
+
+```ruby
+# pick some value
+acts_as_paranoid sentinel_value: DateTime.new(0)
+```
+
+or globally in a rails initializer, e.g. `config/initializer/paranoia.rb`
+
+```ruby
+Paranoia.default_sentinel_value = DateTime.new(0)
+```
+
 ## License
 
 This gem is released under the MIT license.
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -11,14 +11,14 @@ module Paranoia
 
     def with_deleted
       if ActiveRecord::VERSION::STRING >= "4.1"
-        unscope where: paranoia_indexed_column
+        unscope where: paranoia_column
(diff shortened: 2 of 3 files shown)


radar:
I am not sure how this is different from #138. Could you please explain? I think #138 solves the same problem as this.

aaronjensen:
> I am not sure how this is different from #138. Could you please explain? I think #138 solves the same problem as this.

I believe it's quite different for the reason that with the flag you can only have one in each state, one deleted and one not deleted. That means that if you have a unique on `email` and `is_deleted`, and a user who is `foo@bar.com`, if you delete them, then add a new user who is `foo@bar.com`, you can no longer delete them because that would be a unique constraint violation. Simply using a flag does not work. If `is_deleted` were a random number that was > 0, or say, a date, then that would work. But then, you may as well combine them into one column like this does.

smdern:
@aaronjensen raises a really good point #138 opens the possibility of being stuck in a horrible state where if you couldn't delete or restore either model due to unique index constraints.

fphilipe:
> MySQL ignores null columns when it comes to resolving unique index constraints.

I think that this is standard in SQL. You could use a partial index there (not sure if MySQL has this, but PostgreSQL does) such that the uniqueness is only ensured for rows `WHERE is_deleted = false`.

n3bulous:
The boolean state is a bad side effect, but I made the decision not to use magic values as they're generally frowned upon.  @fphilipe's suggestion would be better phrased "Just use PG" :) --  MySQL doesn't support partial indexes in the full sense (you can do partial strings, but not much else).  

The example I've been using sadly demonstrates this limitation of my solution: Users#username.  In retrospect, you shouldn't allow a soft delete of username.  I think this is similar to the email example, and probably a bad practice.

aaronjensen:
While some may frown upon them (I wouldn't agree that they're generally frowned upon, sentinel values are used in all sorts of places), I believe it is the simplest solution besides "use PG". I do not consider the boolean value a solution, since all it does is delay the inevitable. I don't think disallowing soft deletes of certain things is a solution either, since what you're effectively saying is that you can't soft delete anything that has a unique constraint which nullifies the need for a fix.

You could modify the boolean solution to set the value of is_deleted to the id of the item when it was deleted. You couldn't use a tinyint column obviously, but this would solve the problem of deleting things. Of course, it requires two columns to communicate one thing only to workaround a "limitation" of mysql, plus it requires the row to have an id, whereas the sentinel value approach works with the one column. The only thing I don't like about it is that it makes writing sql in the console a little annoying: `deleted_at = '0000-01-01'`

aaronjensen:
One other idea to explore, unique indexes in this problem have the exact opposite semantics that we want. So what if we reverse the meaning of the `is_deleted` column--make it `not_deleted` and its only valid values `1` and `NULL`. That way the unique index on `email`/`not_deleted` will work as we want, enforcing uniqueness on not deleted things and ignoring it on deleted items.

Activerecord may make it difficult to set things to explicitly null, and people may accidentally set it to 0, which would nullify the benefit and it's certainly more confusing.

I wonder if @radar wants to chime in here? I'm concerned that the boolean flag got merged in since it doesn't actually fix the problem and may create more problems for people down the road (they migrate and realize it doesn't actually work and have to migrate to something else)

n3bulous:
Sorry for the long delay.

I think reversing the logic would warrant a separate paranoia fork.  However, I like your idea re: assigning the id to the deleted "flag", which now becomes an integer.  

@radar, what do you think?  I don't think this will change the functionality, just some of the code I added around the boolean.  You haven't done a release yet, so this shouldn't affect anyone but me and my company.

n3bulous:
Bump for opinions. I may actually have a wonderful legacy table that might need less restrictive delete uniqueness.

radar:
I've been wanting to chime in here for a while now, but not sure I have any more opinions to add than those which have already been provided. I'm fine with having a sentinel column on paranoia if that's going to fix a particular bug... A regression test would be a great way of educating me what that bug is. I think I understand from the text description, but I can always understand code a lot better :)

Could you please fix up this PR so that it applies cleanly?

n3bulous:
Thanks for responding @radar. Please see https://github.com/radar/paranoia/pull/150.  I was feeling motivated so I updated my previous code to support the sentinel value.  Not sure if this is exactly what @smdern or @aaronjensen was thinking, but it should handle what they were thinking.  If it doesn't, feel free to remove my code.

aaronjensen:
@n3bulous #150 looks like what I was thinking, it requires 2 columns, but it does fix and not change the semantics of `deleted_at`, which is nice. 

@radar which would you prefer? The bug could be repro'd like this:

1) Create a users table with the columns `username` and `deleted_at`
2) Add a unique index on `username`, `deleted_at`
3) Create two users with the same `username`
4) Notice that that was allowed because of how unique indexes are handled (null values are ignored and considered to not == null)

Now, if you attempted to work around this by using @n3bulous's original patch then:

1) Create a users table with the columns `username`, `deleted_at`, `is_deleted`
2) Add a unique index on `username`, `is_deleted`
3) Create a user with username `foo`.
4) Attempt to create another user with `username` `foo` , this fails, which is good.
5) Delete the user with username `foo`
6) Create another user with username `foo`. This succeeds.
7) Attempt to delete the new user, this fails with a unique index violation, this is bad.

If you'd prefer the single column approach @radar we can clean up this pull, @n3bulous's approach works. It requires two columns, but it doesn't muddy up the meaning of `deleted_at`. It's also annoying to filter out non-deleted things in a sql console using a sentinel of `0000-01-01 00:00:00`. Also, half of it is already there so you wouldn't need to revert the commit.

radar:
I definitely do prefer keeping this simpler. If we could make it one column then that would be best.

aaronjensen:
> I definitely do prefer keeping this simpler. If we could make it one column then that would be best.

This pull request is opt-in and requires only one column. We've also been using it in production for nearly 2 months now.

radar:
@aaronjensen Great! Production usage for 2 months is pretty solid evidence for this being good.

Could you please rebase this so that it applies cleanly to the rails4 branch? Once that happens, I'll merge this in.

aaronjensen:
@radar updated. Note that this also reverts the other incomplete approach.

aaronjensen:
@radar looks like this was cherry-picked in, should we close this one out? Could you drop a new release w/ this please? Thanks!

radar:
@aaronjensen I can close this, but I can't drop a new release. I'm waiting on patches for [the 2.0.3 milestone](https://github.com/radar/paranoia/milestones/Version%202.0.3)

PR #143, opened 2014-07-08
 rails > 4.0.4 + rails 4.1 support

### LINKED ISSUE #129: Rails 4.1 really_destroy! soft deletes the record
Perhaps I'm missing something, however some of my test started failing.

I use paranoia master rails4 branch, I saw `really_deleted!` and tried to give it a go, unfortunately it seems that on Rails 4.1.0 (4-1-stable) doing so just soft deletes the record, that is, it sets the deleted_at value.

not sure if it helps but here's the debug trace:

```
=> 145:       destroy!
   146:     end
   147: 
   148:     include Paranoia
   149:     class_attribute :paranoia_column

(byebug) step

[58, 67] in /Users/kain/.rvm/gems/ruby-2.1.1/bundler/gems/paranoia-d657443970a8/lib/paranoia.rb
   58:   # unless you touch the paranoia column before.
   59:   # We need to override it here otherwise children records might be removed
   60:   # when they shouldn't
   61:   if ActiveRecord::VERSION::STRING >= "4.1"
   62:     def destroy!
=> 63:       destroyed? ? super : destroy || raise(ActiveRecord::RecordNotDestroyed)
   64:     end
   65:   end
   66: 
   67:   def delete

(byebug) destroyed?
false
(byebug) step

[78, 87] in /Users/kain/.rvm/gems/ruby-2.1.1/bundler/gems/paranoia-d657443970a8/lib/paranoia.rb
   78:     end
   79:   end
   80:   alias :restore :restore!
   81: 
   82:   def destroyed?
=> 83:     !!send(paranoia_column)
   84:   end
   85:   alias :deleted? :destroyed?
   86: 
   87:   private

(byebug) n
Next went up a frame because previous frame finished

[58, 67] in /Users/kain/.rvm/gems/ruby-2.1.1/bundler/gems/paranoia-d657443970a8/lib/paranoia.rb
   58:   # unless you touch the paranoia column before.
   59:   # We need to override it here otherwise children records might be removed
   60:   # when they shouldn't
   61:   if ActiveRecord::VERSION::STRING >= "4.1"
   62:     def destroy!
=> 63:       destroyed? ? super : destroy || raise(ActiveRecord::RecordNotDestroyed)
   64:     end
   65:   end
   66: 
   67:   def delete

(byebug) s

[48, 57] in /Users/kain/.rvm/gems/ruby-2.1.1/bundler/gems/paranoia-d657443970a8/lib/paranoia.rb
   48:       end
   49:     end
   50:   end
   51: 
   52:   def destroy
=> 53:     callbacks_result = run_callbacks(:destroy) { touch_paranoia_column(true) }
   54:     callbacks_result ? self : false
   55:   end
   56: 
   57:   # As of Rails 4.1.0 +destroy!+ will no longer remove the record from the db
```

I tried running the paranoia tests locally, and so far just one, possibly not related failure that is also present on travis.

The test for `really_destroy!` passes in the test suite, I also verified the sql queries it is doing and they seems correct, although I believe I'm testing against 4.0.4.

srecnig:
As of rails 4.0.4, touching a column will also fire the after_commit callback. rails changelog: https://github.com/rails/rails/blob/v4.0.4/activerecord/CHANGELOG.md ("Make touch fire the after_commit and after_rollback callbacks.") which means that one test is failing starting with rails 4.0.4.

In this pullrequest i just fixed the test, but I also want to raise the question: wouldn't it be better to work around using `touch`, so that delete no longer raises the `after_commit`callback? This would be nearer on the normal rails behaviour of having `delete` not firing any callbacks. As of now, `delete` also fires the `after_touch` callback, which might be surprising to some people.

I'd be glad to look into it, if people think that's worth doing.

Also, there are two fixes for rails 4.1
- starting with rails 4.1.0 `really_destroy!` only destroys already soft-deleted records, and soft-deletes "normal" records. fixed by calling `touch_paranoia_column` before calling `destroy!`. after calling `really_destroy!` on the dependent objects, i reload the associations on the object, otherwise ActiveRecords's `destroy!` will try to destroy the already really_destroyed associated objects that still exist on the object, but no longer in the database. this fixes #129
- in rails 4.1.2 the reflections have been reworked and `reflections` can no longer be called on activerecord objects, but only on the class (https://github.com/rails/rails/pull/15300). there's still `_reflections`but it's considered private.

Actually there should have been a separate pull request for the rails 4 issue and one for the rails 4.1 fixes, but i messed it up. Sorry, i'm quite new to the whole thing.

 Thanks for your feedback!

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,7 +9,7 @@ platforms :rbx do
   gem 'rubinius-developer_tools'
 end
 
-rails = ENV['RAILS'] || '~> 4.0.2'
+rails = ENV['RAILS'] || '~> 4.1.4'
 
 gem 'rails', rails
 
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -153,8 +153,10 @@ class ActiveRecord::Base
           # Paranoid models will have this method, non-paranoid models will not
           associated_records = associated_records.with_deleted if associated_records.respond_to?(:with_deleted)
           associated_records.each(&:really_destroy!)
+          self.send(name).reload
         end
       end
+      touch_paranoia_column if ActiveRecord::VERSION::STRING >= "4.1"
       destroy!
     end
 
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -144,7 +144,7 @@ class ActiveRecord::Base
     alias :destroy! :destroy
     alias :delete! :delete
     def really_destroy!
-      dependent_reflections = self.reflections.select do |name, reflection|
+      dependent_reflections = self.class.reflections.select do |name, reflection|
         reflection.options[:dependent] == :destroy
       end
       if dependent_reflections.any?
--- a/test/paranoia_test.rb
+++ b/test/paranoia_test.rb
@@ -99,7 +99,7 @@ class ParanoiaTest < test_framework
     assert_equal nil, model.instance_variable_get(:@validate_called)
     assert_equal nil, model.instance_variable_get(:@destroy_callback_called)
     assert_equal nil, model.instance_variable_get(:@after_destroy_callback_called)
-    assert_equal nil, model.instance_variable_get(:@after_commit_callback_called)
+    assert model.instance_variable_get(:@after_commit_callback_called)
   end
 
   def test_destroy_behavior_for_paranoid_models


radar:
Thank you very much @srecnig! I've added this to the rails4 branch now.

PR #148, opened 2014-07-24
include AR callbacks in transaction when running destroy

mtuckergh:
According to https://github.com/rails/rails/blob/4-1-stable/activerecord/lib/active_record/transactions.rb#L78:

```
Both +save+ and +destroy+ come wrapped in a transaction that ensures
that whatever you do in validations or callbacks will happen under its
protected cover
```

acts_as_paranoid overrides the default destroy behavior, but does not currently emulate this transactional wrapping. This PR aims to restore that default AR behavior.

As an example, we have seen issues resulting from this problem manifest with `dependent: destroy` declarations. Sometimes, after a few dependent records are destroyed, something may go wrong and throw an exception before `run_callbacks` is complete and, thus, touch_paranoia_column is never triggered. This leaves the parent record active, but with deleted child records. Since default AR behavior would make this chain of events atomic, I would expect the same behavior with paranoia.

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -56,7 +56,11 @@ module Paranoia
   end
 
   def destroy
-    callbacks_result = run_callbacks(:destroy) { touch_paranoia_column(true) }
+    callbacks_result = transaction do
+      run_callbacks(:destroy) do
+        touch_paranoia_column
+      end
+    end
     callbacks_result ? self : false
   end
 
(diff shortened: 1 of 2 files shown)


radar:
Thanks @mtuckergd! I've added this to the rails4 branch now. Do you think this merits a new paranoia release?

mtuckergh:
I was thinking about that. On one hand, it's more or less a "bug fix". However, it does significantly affect the core behavior of the gem, so I suppose it may be best to increment the rev and push a new version.

mtuckergh:
Been awhile since we touched base on this. Any chance you'll push a new version to RubyGems? There are some projects (ahem, Spree) that could benefit greatly from this fix.
I know you're busy, as am I; just trying not to lose sight of this. :beers:

jamest6550:
@mtuckerd - as I understand this change should allow after_commit callback to fire on destroy when paranoia is installed, correct?

mtuckergh:
@jamest6550 that was not the explicit intent of this change, but I did just verify in a simple rails app that a paranoid delete will trigger the after_commit callback.

This change is intended to bring paranoid behavior more inline with standard ActiveRecord behavior, as described in the PR details.

Specifically, ActiveRecord handles a destroy event as such:

```
-> destroy
  -> start transaction
    -> run callbacks
      -> delete record
```

Prior to this change, Paranoia was acting this way:

```
-> destroy
  -> run callbacks
    -> start transaction
      -> delete record (touch operation)
```

This allowed situations where before_destroy callbacks could (for example) delete child records and an exception occurs somewhere in the batch, but the entire batch is _not_ rolled back because the callback set is not run in a single transaction.

Hope that makes sense.

PR #149, opened 2014-07-28
Added support for restoring records with their own DB connections

hx:
Previously, models with connections different to that of `ActiveRecord::Base` would be restored with a transaction on the wrong connection. Additionally, if there was no connection on ActiveRecord::Base, an exception would be raised.


hx:
Can anyone suggest a solution for the Rails 4.0.\* failure?

radar:
Thanks @hx! I've added this to the rails4 branch.

PR #150, opened 2014-07-28
allow multiple deletes of a unique value

n3bulous:


n3bulous:
Is there a CI config issue with Rails 4.0.8?  The 4.1.4 tests all pass, but the 4.0.8 all have connection issues.

radar:
I believe there has been some issues with recent Rails 4 versions. I have not had time to investigate the severity of these issues yet. I am assuming that they're trivial because nobody is shouting about them.

PR #154, opened 2014-08-25
Reset paranoia_column before updating it. Fixes #153

### LINKED ISSUE #153: "Cannot update a destroyed record" exception on Rails 4.2.0
Rails 4.2.0 will throw an exception when attempting to update columns on a `destroyed?` record:

https://github.com/rails/rails/blob/d20270612cd9be3eed910171934a6b3463dbead7/activerecord/lib/active_record/persistence.rb#L289

This causes restoring of deleted records to fail, since `update_column(paranoia_column, paranoia_sentinel_value)` triggers the exception:

https://github.com/radar/paranoia/blob/rails4/lib/paranoia.rb#L94

woopsn:


lib/paranoia.rb
```diff
@@ -91,6 +91,7 @@ def restore!(opts = {})
         # This only happened on Rails versions earlier than 4.1.
         noop_if_frozen = ActiveRecord.version < Gem::Version.new("4.1")
         if (noop_if_frozen && !@attributes.frozen?) || !noop_if_frozen
+          write_attribute paranoia_column, paranoia_sentinel_value
```
radar, line 94: If we're writing it, why do we then need to update it? Can we get away with just writing it?

woopsn, line 94: `write_attribute` doesn't actually hit the DB, it just assigns the attribute's value. `update_column` is still the call that results in an actual DB update. The problem is that in Rails 4.2, `update_column` will raise an error if `#destroyed?` returns true. So we have to unset `deleted_at` before trying to update the record.

radar, line 94: Ah :) Thank you for clarifying that.

radar:
Thanks @wioux !

woopsn:
Np. Thanks for merging!

PR #161, opened 2014-09-18
Add compatibility info for rails4

dharamgollapudi:


radar:
Thank you @dharamgollapudi :)

PR #165, opened 2014-09-24

shreyas123:
Add a callback called "really_destroy" when the data is actually destroyed as even associations dependency the associated records of paranoid table are soft deleted and only destroyed callback is called. 

Added association with_deleted: true to belongs_to


radar:
I'm confused. Why was this closed?

shreyas123:
@radar I created a new pull request with this feature and added lots of test cases to it.

PR #166, opened 2014-09-25

shreyas123:
As the destroyed callback is called when really_destroyed I needed a callback and ended up adding in the code to call the callback 'really_destroy'

Added association with_deleted: true to belongs_to to fetch all the data including deleted.

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -278,6 +278,6 @@ class ActiveRecord::Base
   end
 end
 
-ActiveRecord::Base.send :include, Paranoia::Association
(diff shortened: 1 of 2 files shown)


radar:
@shreyas123 Is this the PR that you're talking about? It's closed too.

shreyas123:
@radar I am really sorry as I had to make changes in the code before I could make a pull request and I thought this pull request does not made sense thats why I had to close it. I opened another one https://github.com/radar/paranoia/pull/167

shreyas123:
I can remove the with_deleted in belongs to as in Rails 4 I can directly add scope using -> { with_deleted } which does the same work for me as I expect but for the callback I somehow needed it badly

PR #167, opened 2014-09-25

shreyas123:
Added callback really_destroy when the paranoid objects are really destroyed 

Added a association key :with_deleted to belongs_to for activerecord versions >= 4.1

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -278,6 +278,6 @@ class ActiveRecord::Base
   end
 end
 
-ActiveRecord::Base.send :include, Paranoia::Association
+ActiveRecord::Base.send :include, Paranoia::Association if ActiveRecord::VERSION::STRING >= "4.1"
 
 require 'paranoia/rspec' if defined? RSpec
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -17,6 +17,40 @@ module Paranoia
     klazz.extend Callbacks
   end
 
+  module Association
+    def self.included(base)
+      base.extend ClassMethods
+      class << base
+        alias_method_chain :belongs_to, :deleted
+      end
+    end
+
+    module ClassMethods
+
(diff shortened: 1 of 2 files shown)


shreyas123:
@radar I am talking about this one... I added tests for specific to ActiveRecord version >= 4.1 as internally the code is added only if its greater than 4.1

radar:
I am not a fan of all the metaprogramming in this PR. What is this even trying to fix? Perhaps there is a cleaner way of fixing it.

shreyas123:
When soft deleting it calls destroy callback and even when really destroying it calls destroy callback... There is no way to tell when it is actually getting really destroyed.

radar:
Why do you want this feature? In what state of your application do you care that something has been _really_ destroyed?

joshuap:
My use case was cleaning up resources (like uploaded profile pictures) after a hard destroy (removing rhe database record). The record in question was a type of user which could cancel their account (soft-destroy), but the database record was still displayed in certain places in the system for a period of time afterwards.

shreyas123:
There are many places where I needed this callback badly as for polymorphic tables which are not paranoid tables. As there are 2 use cases where a person is marked for deletion (soft-delete) and then really destroyed by the admins.. When really destroyed I want all the associations to be destroyed as well. And as you know if the child is also a paranoid really destroying the parent does not make child really destroyed too. For that case with some checks when parent getting deleted I needed a callback really_destroy to destroy some child or some associations which are not paranoid.

mathieugagne:
I also needed this to remove records from ElasticSearch. I keep all instances (even destroyed ones) and only remove them from the index when they are _really_destroyed_

I can put this in the controller action but it felt right as a callback.

mkou:
yes, I need this feature too

shreyas123:
Closing the pull request as real_destroy callback was added to the latest release

PR #168, opened 2014-09-25
Has one bug

shreyas123:
When the model of has_one association of the paranoid model is not a paranoid then it throws error because of with_deleted called on the association. 

Added a check of paranoid? before executing the query to restore has_one association.


radar:
Could you please add a regression test for this issue too?

shreyas123:
@radar done

jhawthorn:
This test case has been merged and is now passing. Thanks :tada:

PR #170, opened 2014-10-02
Add recovery window option a la acts_as_paranoia

woopsn:
This pull request adds a `:dependent_recovery_window` option to make `paranoia` compatible with that feature of [acts_as_paranoid](https://github.com/ActsAsParanoid/acts_as_paranoid).

When recovering records recursively, associated records will only be restored if they were deleted within the recovery window timespan of the parent's deletion time. This is to enable recovery of a record's associations as they existed at the time it was destroyed.

The default recovery window is set to 2 minutes for compatibility with `acts_as_list`.


sudara:
This should/could/would be merged after #189 — this PR also includes commits from there.

Alternatively we can re-submit a fresh PR after #189 is merged!

Thanks!

mphalliday:
It would be great if we could get this merged, now that #189 has been merged.

bpinto:
:+1:

bluej100:
This would be very helpful to us.

mkou:
:+1:

bluej100:
In my opinion, this should be used not just for restoration but for queries--if you get the children of a deleted record, then by default, it should show you the records which have the same deletion time as the parent. I agree with @radar that it would considerably expand the scope of this project, but I think it's really what people need.

PR #171, opened 2014-10-06
Restoring has_one with recursive: true leads to ActiveRecord::StatementInvalid

srecnig:
i'm using paranoia with rails 4.1.5 and postgres. when trying to restore a record with a has_one association, activerecord will build an invalid query, e.g.

```
SELECT  "authors".* FROM "authors"  WHERE ("authors"."deleted_at" IS NOT NULL) AND (user_id)  ORDER BY "authors"."id" ASC LIMIT 1
```

as far as i understand it, the reason seems to be that the where call in `restore_associated_records` is not given a hash, but rather a string and the id. activerecord interprets this as raw sql and therefore creates the invalid `AND(user_id)`.

changing this to a hash fixes the problem in my app and does not raise any new errors with the test suite. btw, i can reproduce the error only in my app, and not with the test suite that ships with paranoia, i think it's because sqlite does not have a problem with such a query?

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -144,7 +144,7 @@ module Paranoia
       if association_data.nil? && association.macro.to_s == "has_one"
         association_class_name = association.options[:class_name].present? ? association.options[:class_name] : association.name.to_s.camelize
         association_foreign_key = association.options[:foreign_key].present? ? association.options[:foreign_key] : "#{self.class.name.to_s.underscore}_id"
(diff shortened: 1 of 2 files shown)


jhawthorn:
@srecnig Is it possible to add a failing test case? Otherwise this looks good.

srecnig:
@jhawthorn done.

jhawthorn:
Thanks! Excellent test. Makes it clear that this was broken on all databases.

This has been merged as ec04c682bd0a7c4a1669c7f5ffc386914c3941ee. I plan to release a gem including this fix the next day or two.

Might consider setting up travisci to test against all 3 common databases.

PR #177, opened 2014-10-25
Rename Paranoia#destroyed? to Paranoia#paranoia_destroyed?

huoxito:
There was a change in ActiveRecord#update_columns on latest rails
so now it checks for `destroyed?` instead of `persisted?`. Paranoia
can't override `destoyed?` like this anymore otherwise
`update_columns` will always raise:

```
    ActiveRecord::ActiveRecordError: cannot update a destroyed record
```

see rails/rails@0f99aa6

@radar this is a breaking change so not sure how you want to handle it. Couldn't think of a fix other than renaming the method. I'm not sure why `destroyed?` was overridden here in the first place but if tests are missing a scenario this could break something else too. Let you know if I find anything new.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,3 +8,4 @@ rvm:
 env:
   - RAILS='~> 4.0.8'
   - RAILS='~> 4.1.4'
+  - RAILS=master
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -71,7 +71,7 @@ module Paranoia
   # when they shouldn't
   if ActiveRecord::VERSION::STRING >= "4.1"
     def destroy!
-      destroyed? ? super : destroy || raise(ActiveRecord::RecordNotDestroyed)
+      paranoia_destroyed? ? super : destroy || raise(ActiveRecord::RecordNotDestroyed)
     end
   end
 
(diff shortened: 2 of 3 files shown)


radar:
It was being overridden here so that when you soft-deleted a record, `destroyed?` would return true. I'm sad that were probably going to lose that behaviour. I will need to take a look at this patch closer later on.

Thanks mate :)

jhawthorn:
@huoxito @JDutil I'm getting around to this, but it is a pretty major change and I want to be sure nothing else is affected by this change.

This will probably become `2.1.0`, and we'd probably do well to add a warning in the next `2.0.x` release.

radar:
I agree that overriding core methods isn't the way to go and I'm fine with switching it to `#paranoia_destroyed?` with a deprecation in the 2.0.x releases. Please remember to update the README too.

JDutil:
@jhawthorn anything @huoxito or I can do to move this along?

jhawthorn:
@JDutil @huoxito This has been merged into the `rails4` branch, which now tracks the `2.1.x` version. I've just pushed up version `2.1.0.pre`. I'd love for you guys to give this a test and code review.

Issue #169, opened 2014-10-01
NoMethodError - undefined method `to_sym' for nil:NilClass


araslanov-e:
bundle `gem "paranoia", github: "radar/paranoia", branch: "rails4"`

```
User.all

NoMethodError - undefined method `to_sym' for nil:NilClass:
  activerecord (4.1.5) lib/active_record/sanitization.rb:59:in `block in expand_hash_conditions_for_aggregates'
  activerecord (4.1.5) lib/active_record/sanitization.rb:58:in `expand_hash_conditions_for_aggregates'
  activerecord (4.1.5) lib/active_record/relation/query_methods.rb:944:in `build_where'
  activerecord (4.1.5) lib/active_record/relation/query_methods.rb:572:in `where!'
  activerecord (4.1.5) lib/active_record/relation/query_methods.rb:559:in `where'
  activerecord (4.1.5) lib/active_record/querying.rb:10:in `where'
   () .../bundler/gems/paranoia-ba0e7b920239/lib/paranoia.rb:185:in `block in acts_as_paranoid'
  activerecord (4.1.5) lib/active_record/scoping/default.rb:103:in `block (3 levels) in build_default_scope'
  activerecord (4.1.5) lib/active_record/relation.rb:285:in `scoping'
  activerecord (4.1.5) lib/active_record/scoping/default.rb:103:in `block (2 levels) in build_default_scope'
  activerecord (4.1.5) lib/active_record/scoping/default.rb:102:in `block in build_default_scope'
  activerecord (4.1.5) lib/active_record/scoping/default.rb:125:in `evaluate_default_scope'
  activerecord (4.1.5) lib/active_record/scoping/default.rb:101:in `build_default_scope'
  activerecord (4.1.5) lib/active_record/scoping/named.rb:33:in `default_scoped'
  activerecord (4.1.5) lib/active_record/scoping/named.rb:28:in `all'
  activerecord (4.1.5) lib/active_record/querying.rb:10:in `where'
```

radar:
Please provide exact steps to reproduce this issue.

araslanov-e:
Set gem to Gemfile

```
$ bundle
$ bin/rails c
$ => User.all
$ => NoMethodError - undefined method `to_sym' for nil:NilClass
```

all

radar:
I cannot reproduce this problem in a new app. Please upload an app to GitHub which reproduces this problem.

deevis:
[...]

PR #178, opened 2014-11-13
raise error for acts_as_paranoid on models without primary keys

bluej100:
see https://github.com/radar/paranoia/issues/169

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -154,6 +154,7 @@ end
 
 class ActiveRecord::Base
   def self.acts_as_paranoid(options={})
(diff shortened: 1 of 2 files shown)


radar:
Thanks @bluej100!

ndemonner:
This seems to break `db:create` for a vanilla Rails app (it spins up the Rails env, which calls `acts_as_paranoid`, which checks to see if the table has a primary key, but the database doesn't yet exist).

jhawthorn:
This has been reverted due to causing errors when there was no database available.

2ee87601e2b1058b00220c95bd1ff96fe005fdc0

PR #181, opened 2014-11-19
trigger save and create callbacks when restoring

bluej100:
See https://github.com/radar/paranoia/pull/124 .

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -82,15 +82,19 @@ module Paranoia
 
   def restore!(opts = {})
     self.class.transaction do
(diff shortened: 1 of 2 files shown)


radar:
I don't understand the utility of this. The resource is not being created. It's being restored, and I beleive there are callbacks for restoration.

bluej100:
There are, yes, but Rails core and https://github.com/magnusvk/counter_culture have no concept of restore events. As I argue in the linked issue, I do think that restoring is a type of creating, and I think that emitting create events is a better way to work with existing semantics than monkeypatching everything that deals with creation and destruction.

PR #183, opened 2014-11-21
Fix problem when destroying a non-paranoid has_one association

kidlab:
Given these models:

``` ruby
  acts_as_paranoid
  has_one :preference, dependent: :destroy
end

class Preference < ActiveRecord::Base
  belongs_to :user
end
```

It will raise error when performing `really_destroy!` for a `User` record.

``` ruby
user.really_destroy!
#=> NameError: undefined local variable or method `touch_paranoia_column' for #<Preference:0x007fd>
```


morygonzalez:
:+1:

I'm faced with the same problem too.

toshimaru:
same here. :+1:

jhawthorn:
Thanks. This should be fixed in `2.1.0.pre`. :tada:

PR #184, opened 2014-11-23
Add paranoia_scope as named version of default_scope

nyjt:
In the case when you have 2 or more default scopes by 2 or more gems and you unscope the query, but you want to use default_scope of this gem. It is very comfortable if the gem provides a named scope out of box.


jhawthorn:
Seems very reasonable. Thanks @nyjt :star2:

PR #189, opened 2014-12-04
Fix restore with polymorphic has_one relationships

### LINKED ISSUE #174: restore with polymorphic relation fails
When restoring, it search for dependent table's records by the foreign key composed of parent's class name as it does when dealing with simple non-polymorphic relation, eventually fails to find the column and crush.

PatKoperwas:
Restoring polymorphic `has_one` relationships errored because paranoia was
not correctly looking up the foreign_key.

Output from failing test -

```
ActiveRecord::StatementInvalid: SQLite3::SQLException: no such column: parent_model_id: SELECT  "polymorphic_models".* FROM "polymorphic_models" WHERE ("polymorphic_models"."deleted_at" IS NOT NULL) AND (parent_model_id)  ORDER BY "polymorphic_models"."id" ASC LIMIT 1
```

The test sets up a `PolymorphicModel`, which `has_many :parents`. The
`ParentModel` then has a `has_one` relationship with `PolymorphicModel`. When
restoring, the foreign key is set as - `self.class.name.to_s.underscore_id`
which will be `parent_model_id`, instead of the correct `:as` option.

To fix:

If the association object is a `has_one` relationship with an `:as` option,
it will have a `type` attribute (see
https://github.com/rails/docrails/blob/master/activerecord/lib/active_record/reflection.rb#L280).

If this `type` attributes is present, it will be the type column on the
polymorphic model. The foreign key can be found as an attribute on
the association object.

This should fix #174. Solution is based off of @hamidumi 's suggestions in #174.

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -143,8 +143,16 @@ module Paranoia
 
       if association_data.nil? && association.macro.to_s == "has_one"
         association_class_name = association.options[:class_name].present? ? association.options[:class_name] : association.name.to_s.camelize
-        association_foreign_key = association.options[:foreign_key].present? ? association.options[:foreign_key] : "#{self.class.name.to_s.underscore}_id"
-        Object.const_get(association_class_name).only_deleted.where(association_foreign_key => self.id).first.try(:restore, recursive: true)
+        association_foreign_key = association.foreign_key.present? ? association.foreign_key : "#{self.class.name.to_s.underscore}_id"
+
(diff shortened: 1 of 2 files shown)


sudara:
Let me know if there's anything I can do to help get this beauty merged! :)

emilyadavis303:
Any word on when this might get merged in?

jhawthorn:
This was merged. Thanks very much. :cake:

Issue #28, opened 2012-07-28
Including some built-in RSpec shared examples


jcf:
Would you be interested in including a shared example to make testing use of `acts_as_paranoid` more simple?

I use the following shared example in my apps to make sure the default scope has been modified as expected.

``` ruby
shared_examples_for 'a paranoid model' do
  describe 'acts_as_paranoid' do
    context 'with the default scope' do
      it 'adds a deleted_at call to all selects' do
        described_class.scoped.where_values_hash.
          should include(deleted_at: nil)
      end
    end

    context 'when unscoped' do
      it 'skips adding the deleted_at condition' do
        described_class.unscoped.where_values_hash.
          should_not have_key(:deleted_at)
      end
    end
  end
end
```

Irio:
:+1:

radar:
@Irio Yes I would be interested in this. Could you please build a PR for this?

radar:
Closing due to staleness.

joshuapinter:
I know you closed this but what's the best method to test that acts_as_paranoid is being included on a model, in Rspec?

`described_class.scoped.where_values_hash` results in `{}`. 

Cheers!

PR #195, opened 2014-12-19
Add a simple Testing with Rspec section.

joshuapinter:
A simple shared examples for Rspec that helps people easily add tests to ensure that their models are properly including the `acts_as_paranoid` library and the model's table is properly setup with a `deleted_at` column that is indexed.

_I know you're huge on documentation and your docs for Paranoia are some of the cleanest and understandable I've seen. So, I wanted to include this to help others easily test their use of Paranoia with Rails and Rspec. Also, this might only work for SQL-based databases because it calls `where_sql` and is syntax dependent, but I think it's useful none-the-less._

_Cheers._

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,7 +205,7 @@ Paranoia.default_sentinel_value = DateTime.new(0)
 
 ## Testing with Rspec
 
-You can use shared examples to help test that models are including `acts_as_paranoid` properly. For example, the following shared examples can be created in `spec/support/shared_examples/paranoid_examples.rb`:
+You can use shared examples to help test that models are including `acts_as_paranoid` properly. For example, the following shared examples can be created in `spec/support/shared_examples/paranoia_examples.rb`:
 
 ```ruby
 shared_examples_for 'a Paranoid model' do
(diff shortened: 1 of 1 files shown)


joshuapinter:
Replaces Issue #28.

joshuapinter:
By the way, I discovered that adding `acts_as_paranoid` to an abstract base class will not yield the proper query, but will include a blank table name instead of the proper name. I created PR #196 to add some documentation around that for others. 

I also modified one of my Rspec examples to check for this. Not sure if this is too DB-dependent but it looks like this now: 

``` ruby
it 'adds a deleted_at where clause' do
  table = described_class.to_s.underscore.pluralize
  described_class.scoped.where_sql.should include("`#{table}`.deleted_at IS NULL")
end
```

jozr:
:+1:

jhawthorn:
Thanks @joshuapinter I added this to the wiki here: https://github.com/radar/paranoia/wiki/Testing-with-rspec

Do you think any of this should be moved to our `:acts_as_paranoid` matcher? https://github.com/radar/paranoia/blob/rails4/lib/paranoia/rspec.rb

joshuapinter:
@jhawthorn Looks great! 

It certainly would be useful to check for more than `Paranoia` being included by the model's ancestors. However, if we're gonna included any checks on the columns, we should abstract out `deleted_at` to be able to pull in whatever is set for the `column` option.

I think we leave it for now unless somebody wants to do that, the wiki is helpful as it is. 

:+1:

PR #196, opened 2014-12-21
Add section for inheriting from an abstract class.

joshuapinter:
An issue arises when inheriting from an abstract class and the `acts_as_paranoid` is used on the abstract class instead of the class inheriting from the base class.

This just produces a query with an empty table name and fails fairly silently. 

Not sure if this is the best place for it but it's important for users to know about.


jhawthorn:
Thanks @joshuapinter. Added to the wiki https://github.com/radar/paranoia/wiki/Inheriting-from-an-Abstract-Class

joshuapinter:
:+1:

PR #156, opened 2014-08-28
Fix RSpec 3.x deprecations

lime:
The `failure_message_for_should` syntax is deprecated in RSpec 3.

This could of course be merged into the `rails3` branch as well. Do you want me to make a separate pull request for that?

PR #197, opened 2014-12-23
Fix matcher compatibility with RSpec 2

lime:
I messed up with #156. I was thinking of the other RSpec matcher syntax where you define `failure_message_when_negated` as a method, rather than use it as a DSL call...

This way of retaining compatibility with both RSpec 2 & 3 is taken straight from the [corresponding file in the `pundit` gem](https://github.com/elabs/pundit/blob/v0.3.0/lib/pundit/rspec.rb).


jhawthorn:
Thanks

PR #205, opened 2015-01-26
Trace feature

mareklipka:
This feature allows to keep track of who deleted specific record. By default, it's enough to add `deleted_by_id` column to appropriate table, use `acts_as_paranoid trackable: true` in model and pass `deleted_by_id` as an option to `destroy` method.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,6 +168,56 @@ If you want callbacks to trigger before a restore:
 before_restore :callback_name_goes_here
 ```
 
+#### Tracking
+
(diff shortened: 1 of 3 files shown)


jhawthorn:
Hi thanks for the PR, but I think this feature is outside the scope of paranoia.

It would be better to implement this in your own app, preferable with a method for that purpose, something like:

``` ruby
def destroy_by_user(user)
  update! destroyed_by: user
  destroy
end
```

radar:
:+1: @jhawthorn

PR #210, opened 2015-02-10
fix ActiveRecord::StatementInvalid: PG::UndefinedTable (https://github.com/rails/rails/issues/12770)

vlewin:
fix ActiveRecord::StatementInvalid: PG::UndefinedTable: invalid reference to FROM-clause ..., use table name alias (https://github.com/rails/rails/issues/12770)

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -182,7 +182,7 @@ class ActiveRecord::Base
     self.paranoia_column = (options[:column] || :deleted_at).to_s
     self.paranoia_sentinel_value = options.fetch(:sentinel_value) { Paranoia.default_sentinel_value }
     def self.paranoia_scope
-      where(table_name => { paranoia_column => paranoia_sentinel_value })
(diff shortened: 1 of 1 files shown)


vlewin:
This is still an issue in 4-0-STABLE and 4-1-STABLE for me.

jhawthorn:
I'm unsure what the issue is. Could you provide a test which fails without this patch?

vlewin:
@jhawthorn Hi John, thank you for quick response. Here is the test which fails without the patch:
https://gist.github.com/vlewin/6836f490c94a126f0fc4

Changing gem 'paranoia' to gem 'paranoia', github: 'vlewin/paranoia', branch: 'rails4' in the Gemfile will make the test green again.

vlewin:
The issue here is the hard coded table name instead of table alias which will be used in :build_default_scope method.

vlewin:
Marked as a duplicate of this pull request #212

PR #223, opened 2015-04-08
Update README.md

akshaysmurthy:
Show the `restore` instance method also in the code snippets
- `client.restore`
- `client.restore(:recursive => true)`


radar:
Thanks!

PR #227, opened 2015-04-14
check paranoia_column existence

jguitar:
We were using Rails models on database migrations, so if we run `rake db:migrate` from scratch will fail as 

> rake aborted!
> StandardError: An error has occurred, this and all later migrations canceled
> 
> SQLite3::SQLException: no such column: posts.deleted_at: SELECT "posts".\* FROM "posts" WHERE "posts"."deleted_at" IS NULL/Users/juanfperez/.rvm/gems/ruby-head/gems/sqlite3-1.3.10/lib/sqlite3/database.rb:91:in `initialize'

This can be solved checking for the column existence.

_Of course, it can be solved if we made a trick on the migration file. That way we can avoid the call to `act_as_paranoid`_

``` ruby
class MigrateData < ActiveRecord::Migration
  class Post < ActiveRecord::Base;end

  def change
    Post.all.each do |post|
      post.update_attributes title: 'untitled'
    end
  end
end
```

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -201,7 +201,7 @@ class ActiveRecord::Base
     def self.paranoia_scope
       where(paranoia_column => paranoia_sentinel_value)
     end
-    default_scope { paranoia_scope }
+    default_scope { paranoia_scope if columns_hash.has_key? paranoia_column }
 
     before_restore {
       self.class.notify_observers(:before_restore, self) if self.class.respond_to?(:notify_observers)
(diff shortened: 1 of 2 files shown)


radar:
Don't reference models in migrations. It is a code smell and a hack to re-define the model like that. Instead:

``` ruby
connection.execute "UPDATE posts set title = 'untitled'"
```

jguitar:
Thanks @radar for the answer. I know I should use SQL, but sometimes there is a lot of logic to be done on the massive updates. So I used the hack.

PR #228, opened 2015-04-21
Failing test case for "after_commit" on destroy

skorfmann:
We're just upgrading an application which uses the paranoia gem (rails4 branch) to Rails 4.2.1 - We've several models which are using conditional `after_commit` hooks, e.g.: 

``` ruby
after_commit   -> { do_stuff }, on: :destroy
```

This breaks for `destroy`, since it's [depending on](https://github.com/rails/rails/blob/c539cc006179c98d418c6914ce567f657b6f1966/activerecord/lib/active_record/transactions.rb#L433-L444) `destroy?` which changed its behaviour with #177. 

Our workaround will probably look somehow like this:

``` ruby

class ActiveRecord::Base
  private

  # Determine if a transaction included an action for :create, :update, or :destroy. Used in filtering callbacks.
  def transaction_include_any_action?(actions) #:nodoc:
    actions.any? do |action|
      case action
      when :create
        transaction_record_state(:new_record)
      when :destroy
        destroyed? || (respond_to?(:paranoia_destroyed?) && paranoia_destroyed?)
      when :update
        !(transaction_record_state(:new_record) || destroyed? || (respond_to?(:paranoia_destroyed?) && paranoia_destroyed?))
      end
    end
  end
end
```

However, we're wondering if anyone has (or could think of) a nicer solution for this?

### DIFF
--- a/test/paranoia_test.rb
+++ b/test/paranoia_test.rb
@@ -111,6 +111,14 @@ class ParanoiaTest < test_framework
     assert model.instance_variable_get(:@after_commit_callback_called)
   end
 
+  def test_destroy_behavior_for_conditional_models_callbacks
+    model = ConditionalCallbackModel.new
+    model.save
+    model.remove_called_variables     # clear called callback flags
+    model.destroy
+
+    assert model.instance_variable_get(:@after_commit_callback_called)
+  end
 
   def test_delete_behavior_for_plain_models_callbacks
     model = CallbackModel.new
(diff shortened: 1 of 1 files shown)


Martin288:
@skorfmann I am facing the same problem because our application is using paranoia and sunspot together. The latter is dependent on the `after_commit :hook, on: :create` callback but paranoia has broken it. I wonder if I can take your monkey patching as a working solution.

skorfmann:
@Martin288 I think we went this solution and it's working fine for us

Martin288:
@skorfmann cool! But I have fallback to use `:after_destory` instead_of `:after_commint` for sunspot's `auto_remove_callback` configuration now. And it also works for our case. The README has some related tips: https://github.com/rubysherpas/paranoia#callbacks. Hope paranoia will fix this ASAP.

MatayoshiMariano:
:+1: to this PR please!

AlexRiedler:
I am not exactly clear why but the workaround @skorfmann didn't work for me until I changed `transaction` in `#destroy` to be `with_transaction_returning_status` instead; otherwise the attributes are not set on the model when doing the `transaction_include_any_action` check; and as a result would not run the necessary callbacks.

^ not sure if the above is necessary as well for this change; but might give insight for anyone working on it.

Also this test given might only works in Rails 5 since `after_commit` are not called in a lot of rails test environments (rails 4 and less need something like `test_after_commit` gem).

PR #229, opened 2015-04-22
Recovery window

bluej100:
rebase of https://github.com/radar/paranoia/pull/170


radar:
I would like to keep paranoia simple and I think by adding this feature, it is not going to keep it that way. If you would like this feature please maintain it on your own fork of paranoia.

PR #230, opened 2015-04-22
Add ability to have types other than datetime and add boolean type.

aselder:
This allows an optional :type option to be passed to acts_as_paranoid.
The allowed types are :datetime (the default), and :boolean. If the
column type is boolean, true will be written to the attribute rather
than the date/time.

It should follow all the same behavior as the datetime version, except
for callbacks. The default implementation uses AR::Base#touch which
fires after_touch, after_commit, and after_rollback callbacks. The
boolean version uses update_column which fires no call backs. I
couldn't find an equivalent method for a non-datetime field.

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -1,6 +1,8 @@
 require 'active_record' unless defined? ActiveRecord
 
 module Paranoia
+  KNOWN_PARANOIA_TYPES = [:datetime, :boolean]
+  
   @@default_sentinel_value = nil
 
   # Change default_sentinel_value in a rails initilizer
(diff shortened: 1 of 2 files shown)


radar:
What is the point of this? In what case would you want to have a boolean? How is this useful for others who use this library?

aselder:
We have existing data schemas that use a boolean column (is_deleted) to denote soft deleted records rather than a datetime column. This allows user to use a column of some other type rather than only a date time

radar:
Run a migration which converts the booleans to datetimes, where the time is the current time.

Alternatively, maintain a fork.

aselder:
Remember, not everybody has control of their schemas. I'll just maintain a fork, but thought this ability might be useful to others.

PR #232, opened 2015-05-09
partial indexes advice

levraipixel:
It took me a few hours to figure that out, so I guess it can help other people save time


levraipixel:
thanks

PR #234, opened 2015-05-18
Paranoia.without!

ksz2k:
It is sometimes needed to run some code without Paranoia's default_scope, so here is the patch for it :-) 

Eg.

``` ruby
Paranoia.without!
User.staff.find_each do |user|
  # do some work here
end
```

or even better:

``` ruby
Paranoia.without do
  User.staff.find_each do |user|
    # do some work here
  end
end
```

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -2,7 +2,7 @@ require 'active_record' unless defined? ActiveRecord
 
 module Paranoia
   @@default_sentinel_value = nil
-  @@disabled = false
+  @@without = false
 
   # Change default_sentinel_value in a rails initilizer
   def self.default_sentinel_value=(val)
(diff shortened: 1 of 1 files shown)


radar:
Does Model.unscoped work or is the default_scope still applied?

ksz2k:
I had exceptions eg when there was model used in migrations (`User.staff.find_each`), before `deleted_at` column was added.. adding unscoped (`User.unscoped.staff.find_each`) didn't help..

radar:
Can you avoid using the model in the migrations?

ksz2k:
Sure :-) Won't do next time.. 
But I'm still on project where these migrations already are..

jhawthorn:
This isn't threadsafe, so we cannot do this.

Even if it were fixed to use `Thread.current`, I think it's dangerous to have a global state control how paranoid works. It's also unclear what this does: `Paranoid.without{ user.destroy }` will still soft-delete records.

So I am :-1:

ksz2k:
Ok, thanks for feedback

PR #245, opened 2015-06-16
Introduce #paranoia_destroy_attributes and #paranoia_restore_attributes extension points

Empact:
We maintain a second `active` column for the sake of enforcing uniqueness among un-deleted rows - this is a sort of more general solution to the same problem https://github.com/radar/paranoia/wiki/Custom-sentinel-values attempts to solve (incidentally, a problem more elegantly solved by http://www.postgresql.org/docs/9.4/static/indexes-expressional.html, if one has access to them).

As an alternative to maintaining this extra value via a separate callback, it would be nice to update it at the same time paranoia updates the `deleted_at` column. We can do that and perhaps serve other needs by introducing these 2 extension points. With them, getting the behavior we need is a simple matter of adding a value to each of the hashes from these methods.

One significant difference brought on by this change is that `#touch` occurs inside a transaction, whereas `#update_columns` creates no such transaction. This only affects `#delete` because `#destroy` already creates its own transaction. To my eye, this behavior is more in line with the activerecord documentation and default behavior, which makes no mention of transactions for `#delete` but it is a change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ If you wish to actually destroy an object you may call `really_destroy!`. **WARN
 If a record has `has_many` associations defined AND those associations have `dependent: :destroy` set on them, then they will also be soft-deleted if `acts_as_paranoid` is set,  otherwise the normal destroy will be called.
 
 ## Getting Started Video
-Setup and basic usage of the paranoia gem  
+Setup and basic usage of the paranoia gem
 [GoRails #41](https://gorails.com/episodes/soft-delete-with-paranoia)
 
 ## Installation & Usage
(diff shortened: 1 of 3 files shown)


Empact:
The alternative we're taking advantage of at the moment is to override the behavior of `touch_paranoia_column` and `restore!`, which is more duplicative than I would like.

Empact:
Failure looks spurious to me.

radar:
Looks good to me. Could you please add some documentation for this to the README?

Empact:
I updated the branch here: https://github.com/brigade/paranoia/tree/update_attributes doesn't seem to be showing up in the diff here. Changes are that I update the readme, add some testing, and fixed an issue in .only_deleted, which would miss records when a null value signifies deletion.

Empact:
@radar I think this is ready to go

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # paranoia Changelog
 
+## Unreleased
```
radar: I've done a bad job of keeping this up to date. Normally what I would do is just point people at the changes between tags.

README.md
```diff
@@ -207,6 +207,49 @@ add_index :clients, [:group_id, :other_id], where: "deleted_at IS NULL"
 
 Of course, this is not necessary for the indexes you always use in association with `with_deleted` or `only_deleted`.
 
+##### Unique Indexes
+
+Becuse NULL != NULL in standard SQL, we can not simply create a unique index
+on the deleted_at column and expect it to enforce that there only be one record
+with a certain combination of values.
+
+If your database supports g, good alternatives include partial indexes
```
radar: "supports g" <- is that right?

lib/paranoia.rb
```diff
@@ -29,7 +29,16 @@ def with_deleted
     end
 
     def only_deleted
-      with_deleted.where.not(paranoia_column => paranoia_sentinel_value)
+      if paranoia_sentinel_value == nil
```
radar: Prefer `paranoia_sentinel_value.nil?` here.

radar:
LGTM aside from the comments I've left.

CHANGELOG.md
Empact: Would you rather I drop it then?

radar: Yup, that'd be preferable.

Empact: :+1:

radar:
This PR does not merge cleanly to master. Please update.

Empact:
@radar Done

radar:
Thanks!

Issue #175, opened 2014-10-22
Deleted association query don't work


hamidumi:
If i call association method on deleted instance(association deleted too), association search in undeleted records.

radar:
Fixed by #122.

PR #247, opened 2015-06-21
Fix for #122, #121 and #175 on Rails 4 branch

codeodor:
I wasn't sure whether #121 got closed because this fix wasn't desired for the Rails 4 branch, or just because it was so similar to #122 that it looked like a duplicate, so while I was fixing [the issue brought up](https://github.com/radar/paranoia/pull/122#issuecomment-113934008) by @joshuapinter I thought I'd add the test for the Rails 4 branch as well.

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -237,3 +237,16 @@ class ActiveRecord::Base
 end
 
 require 'paranoia/rspec' if defined? RSpec
+
+module ActiveRecord
+  module Validations
+    class UniquenessValidator < ActiveModel::EachValidator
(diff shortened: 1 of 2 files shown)


codeodor:
For what it's worth, this test that's failing on travis fails before this change. If I `git checkout HEAD~3` from where it's at now, and run against jruby, it results in the same failure.

codeodor:
But I fixed it anyway. That last commit should do it.

radar:
Thanks @codeodor!

codeodor:
My pleasure! :+1:

Issue #250, opened 2015-06-30
Problem with latest commit


marcoschicote:
Hi
The latest commits are breaking my tests. I was already using paranoia 2.1.3 and I had to run `bundle update` to update other libraries and I got a revision change in paranoia. Previous revision was  3f72776cee2c5e5b9aef7a5e4a4e0e4ff9c18eb4 and after update I got revision 8a805cd2dcd00191d2ba4a22fc494a06f873607b. This caused my tests to start failing with the following message: 

`NoMethodError: undefined method `paranoia_column' for #<Class:0x007ff647691280>`

I'm a little out of time as to produce a minimum example for this but wanted to let you guys now. 
Hope this helps somehow. I apologize for the little information. 

Marcos

BattleBrisket:
:+1: Seeing the same behavior. 

However I was pulling directly from the github repo, so I guess I can only blame myself. Swapping to the normal gem publication removed the error for me.

```
gem "paranoia", "~> 2.0"
```

radar:
Please provide a stacktrace for the issue so I can track it down. Steps to reproduce would also be appreciated.

radar:
I think this was fixed by #251.

mlabarca:
Hi, I think this issue persists in the rails3 branch. Creating any model instance not related to paranoia seems broken due to uniqueness validator attempting to access a non existant paranoia column. Could  the fix for #251 (3b6ff6ec) be merged into the rails 3 branch? overriding the build_relation_with_paranoia method in an initializer with the same changes fixed the issue for me using the latest version from rails 3 branch.

radar:
Please submit a PR to fix this issue and I will review.

PR #251, opened 2015-07-01
Fix uniqueness validator for non-paranoid associations

milgner:
One of my paranoid models has a relation to a non-paranoid sub-model. With the current version of the `UniquenessValidator` enhancement, this caused paranoia to fail.
I added a test for this constellation and implemented the corresponding fix.


raphaelcosta:
I fixed the same issue at same time =/ 
https://github.com/raphaelcosta/paranoia/commit/37861586bef4fae84f5ae284febd5b06ef2def77

radar:
I believe this is a fix for #250. Thank you!

PR #256, opened 2015-07-25
Fix uniqueness validator for non-paranoid models.

mlabarca:
As I mentioned on #250, simply attempting to create a non paranoid model with an uniqueness validation fails since it attempts to filter out records with paranoid column != nil from the relation. This is the very same fix in 3b6ff6ec but for this rails3 branch which is what I'm depending on.

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -181,7 +181,11 @@ module ActiveRecord
       protected
       def build_relation_with_paranoia(klass, table, attribute, value)
         relation = build_relation_without_paranoia(klass, table, attribute, value)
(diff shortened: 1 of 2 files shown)


leklund:
:+1: for fixing this for the rails3 branch.

It could also be done guard clause style since that's how it is the rails4 branch currently: https://github.com/rubysherpas/paranoia/blob/rails4/lib/paranoia.rb#L265

leklund:
After some more thought, I'm in favor of reverting the uniqueness validator as in #277

Jesus:
Many thanks for this patch.

I had to fork the repo in order to use because it looks like it won't be merged. But that's ok for me.

PR #257, opened 2015-07-31
Update paranoia.gemspec

OleOle7177:
Update activerecord dependency to work with later activerecord version: for use paranoia with rails 5, that depends on activerecord ~> 5.0


radar:
Nope.

radar:
http://stackoverflow.com/questions/5170547/what-does-tilde-greater-than-mean-in-ruby-gem-dependencies

radar:
AR 5 could have breaking changes. There is no way that I will be merging a PR like this at all. Ever.

Issue #258, opened 2015-08-03
Uniqueness validator does not work well with new paranoia_sentinel_value


vaharoni:
If I follow the README and do this in my model:

``` ruby
validates_uniqueness_of :email
acts_as_paranoid column: :active, sentinel_value: true
def paranoia_restore_attributes
  {
    deleted_at: nil,
    active: true
  }
end
def paranoia_destroy_attributes
  {
    deleted_at: current_time_from_proper_timezone,
    active: false
  }
end
```

Then running `x.valid?` on that object in console performs the following validation query:

``` sql
SELECT  1 AS one FROM "users" WHERE ("users"."email" = 'email@email.com' AND "users"."active" IS NULL) LIMIT 1
```

which is not desired, as we want `"user"."active" = 't'`.

I noticed in the code the monkey patching of `UniquenessValidator` that looks like this:

``` ruby
if klass.respond_to?(:paranoia_column)
  relation.and(klass.arel_table[klass.paranoia_column].eq(nil))
else
  relation
end
```

Shouldn't `eq(nil)` be changed to `eq(klass.paranoia_sentinel_value)` ?

UPDATE:
I created PR to solve the issue.

radar:
Fixed by #259.

PR #259, opened 2015-08-03
Fixed issue #258: uniqueness validator now works with custom active c…

vaharoni:
…olumn


radar:
Yes you're right, it should use the sentinel value instead. Thanks for the PR and tests :)

PR #266, opened 2015-09-11
Not deleted scope

matfiz:
Hi Radar,

In this PR I add an additional scope `not_deleted` that mimic the default_scope provided by this gem. 

I considered it a usefult feature in the following scenario. There is a model `TestModel` with a complex `default_scope`. On a particular type of queries, when you call `TestModel.unscoped` then you get rid also of the default_scope provided by paranoia. In order to provide a descriptive scope, now it is possible to call `TestModel.unscoped.not_deleted`

I hope you like it :)

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -41,6 +41,10 @@ module Paranoia
       end
     end
     alias :deleted :only_deleted
+    
+    def not_deleted
+      where(paranoia_column => paranoia_sentinel_value)
+    end
 
     def restore(id_or_ids, opts = {})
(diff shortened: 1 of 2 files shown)


jhawthorn:
I believe this is the same as `.paranoia_scope`, perhaps it should be an `alias_method` of that.

rbr:
Heh, I didn't see this before, but that's exactly what was added in https://github.com/rubysherpas/paranoia/pull/305.

matfiz:
@jhawthorn @rbr Ok, then I guess this PR should be closed :) Thanks!

PR #267, opened 2015-09-15
Using guard clauses.

diego-aslz:
[Prefer Guard Clauses](http://www.thechrisoshow.com/2009/02/16/using-guard-clauses-in-your-ruby-code). Hope you like it :smiley:


radar:
Yup, I like it :)

Issue #264, opened 2015-09-01
[question] object.restore doesn't touch updated_at column


abitdodgy:
It appears that `restore` method doesn't change the `updated_at` timestamp. This is probably by design, so I would like to understand the reasoning.

Potentially, this is an issue when using fragment caching. For example, an object can be destroyed and then rendered and cached in a separate view. After it's restored the cache fragment will be from its destroyed state.

It's very easy to work around this, but I'm just curious.

radar:
I agree that it should change the updated_at time stamp. Please submit a fix for this.

abitdodgy:
Sounds good! I'll send one!

abitdodgy:
I still plan on issuing a pull request for this issue. I just haven't had a chance to yet.

abitdodgy:
It looks like @leonardoalifraco beat me to it.

PR #269, opened 2015-09-26
Touching record when restored.

leonardoalifraco:
According to what was discussed on #264 I've added the corresponding touch when the record is restored.

### DIFF
--- a/test/paranoia_test.rb
+++ b/test/paranoia_test.rb
@@ -765,7 +765,7 @@ class ParanoiaTest < test_framework
     pt1.update_columns(created_at: 20.years.ago, updated_at: 10.years.ago, deleted_at: 10.years.ago) 
     ParanoidModelWithTimestamp.record_timestamps = true
     assert pt1.updated_at < 10.minutes.ago
-    assert !pt1.deleted_at.nil?
+    refute pt1.deleted_at.nil?
     pt1.restore!
     assert pt1.deleted_at.nil?
     assert pt1.updated_at > 10.minutes.ago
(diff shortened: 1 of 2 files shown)


abitdodgy:
@leonardoalifraco I see you've tried to take care of issue #264. Thanks for taking care of that. An aside, it's a good idea to check if someone has started work on something they have take ownership of before taking it on.

I'm curious, isn't there a better way than to use touch? When restoring we are already updating the `deleted_at` column, so why not take this opportunity and update the `updated_at` too instead of using `touch`, which, from what I understand, issues a separate query?

test/paranoia_test.rb
```diff
@@ -757,6 +758,19 @@ def test_validates_uniqueness_still_works_on_non_deleted_records
...
   end
 
+  def test_updated_at_modification_on_restore
+    parent1 = ParentModel.create
+    pt1 = ParanoidModelWithTimestamp.create(:parent_model => parent1)
+    ParanoidModelWithTimestamp.record_timestamps = false
+    pt1.update_columns(created_at: 20.years.ago, updated_at: 10.years.ago, deleted_at: 10.years.ago) 
+    ParanoidModelWithTimestamp.record_timestamps = true
+    assert pt1.updated_at < 10.minutes.ago
+    assert !pt1.deleted_at.nil?
```
radar: I would use a `refute` here. I think that would work.

radar:
LGTM besides that one comment. Let me know if the `refute` works or not and then I'll merge.

lalifraco-devspark:
@abitdodgy Sorry to takeover this issue. But I was  investigating this gem and decided to start contributing with something that was already confirmed and asked for a change.
Regarding to the implementation details, it's making 2 different database queries, but the AR touch implementation encapsulates the modification of the updated_at/on column functionality and from my perspective the clearer path to address this would be to keep it DRY rather than copying the code.

@radar Great. I agree that the refute is clearer. I'll make the change.

leonardoalifraco:
@radar I've already changed the assertion strategy using refute.

radar:
Thanks @leonardoalifraco. I've been busy :)

PR #270, opened 2015-10-16
Ensure delete_all and destroy_all are paranoid

unimatrixZxero:
- Previously: delete_all calls would simply erase the models form the database 
- Previously: there weren't any specs for destroy_all

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -249,6 +249,10 @@ class ActiveRecord::Base
   def self.paranoid? ; false ; end
   def paranoid? ; self.class.paranoid? ; end
 
+  def self.delete_all(conditions=nil)
+    super and return unless paranoid?
+    where(conditions).update_all(self.new.send :paranoia_destroy_attributes)
+  end
   private
 
   def paranoia_column
--- a/test/paranoia_test.rb
+++ b/test/paranoia_test.rb
@@ -879,6 +879,36 @@ class ParanoiaTest < test_framework
     related.valid?
   end
 
+  def test_delete_all
+    model = ParanoidModel
+    count = 3
+    models = count.times.map { model.new }
+    assert_equal 0, model.count
+    models.each(&:save!)
+
+    assert_equal count, model.count
+    model.delete_all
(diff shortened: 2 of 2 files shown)


radar:
I actually prefer this behaviour as it is now. `delete_all` _should_ act fine. If you want the destruction of those objects to be stopped, use `destroy_all` instead.

unimatrixZxero:
@radar By that logic, we should also remove the `#delete` override!

ivan-kolmychek:
Current behavior is kind of incosistent, yes. Shouldn't `delete_all` and `delete` behave similarly?

unimatrixZxero:
@ivan-kolmychek Yes, exactly why I proposed this PR.

radar:
What is the usecase of this method? Why would you want to paranoia-delete-all of your records at once? I can understand the usecase of doing destroy + delete in a paranoia fashion (otherwise, why would've I written this library?) but I don't understand why you'd want to `delete_all`.

Is it mass-deleting of objects? Is that it?

unimatrixZxero:
Well if you are operating on some scope it would be nice to be able to soft-delete those too:  e.g. `SupportTickets.older_than_a_year_scope.delete_all`(just a made up example)

krainboltgreene:
I can definitely see how it would be nice to not have to look at all of your queries that use `delete_all` after introducing paranoia, and just have them _work_.

lib/paranoia.rb
```diff
@@ -249,6 +249,10 @@ def self.I_AM_THE_DESTROYER!
   def self.paranoid? ; false ; end
   def paranoid? ; self.class.paranoid? ; end
 
+  def self.delete_all(conditions=nil)
+    super and return unless paranoid?
```
caiguanhao, line 253: This is a bug. Now Model.delete_all always returns nil but not a number. It should be `return super unless paranoid?`.

timherby:
This would be very useful.  I see it was merged into `rails4` but not the main branch.  Any chance we can get this merged into the core gem?

PR #283, opened 2015-11-30
Adding support to cancel "restore" and "real_destroy" actions

arieliten:
Currently the callbacks for `before_restore` and `before_real_destroy` can't be canceled (via returning `false`) like any other normal ActiveModel::Callback. So here I'm adding the support to cancel any of those actions (`restore` or `real_destroy`) by returning `false` in their callbacks.

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -55,7 +55,7 @@ module Paranoia
   module Callbacks
     def self.extended(klazz)
       [:restore, :real_destroy].each do |callback_name|
-        klazz.define_callbacks callback_name
+        klazz.define_callbacks callback_name, terminator: ->(_, result) { result == false }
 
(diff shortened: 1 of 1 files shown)


rbr:
Canceling callbacks with `false`seems to be deprecated in Rails 5: https://github.com/rails/rails/pull/17227, `throw(:abort)` is the new way.

valentindruon:
What if we are not using Rails 5.0?

therrick:
I'd love to see this get merged. We've been using a custom fork with this change for several months now and would much rather have it in the main repo. Any outlook for accepting it?

BenMorganIO:
Good job @arieliten. Can you provide a test for this change?

kasparsj:
+1

PR #284, opened 2015-12-02
Rails5

gagalago:
This is the change mandatory for rails5. I have not updated the version of paranoia because i don't know your rules about that.


gagalago:
I have found an error on UniquenessValidator on line 265.

```
undefined method `and' for #<ActiveRecord::Relation []>
```

but I don't understand it.

jhawthorn:
Thanks for the work! This was cherry-picked with modifications into master, which should now support rails 5.

PR #287, opened 2015-12-15
Paranoia bug combined with arel-helpers

mtparet:
Hey,

I discovered a bug using Paranoia and Arel and wrote a test showing it.
The `deleted_at` is not correctly set up in the SQL generated. (cf comment on failing test)


krainboltgreene:
@mtparet Can you repush so we can get travis running this?

mtparet:
Bug opened also on arel helpers side https://github.com/camertron/arel-helpers/issues/22

PR #290, opened 2015-12-21
Implement Rails 5.0 compatibility

halostatue:


halostatue:
Note that this includes #286 because of the regression involved, plus this is something that was ultimately commented out of the other Rails 5 branch.

seanhandley:
Note that Rails needs Ruby 2.2.0 or greater, hence the Travis failures here.

See https://github.com/rails/rails/commit/32f7491808d2c4e097ed7b3ee875e4d1cea8c442#diff-73ac7236f7a8247a9cd61d338ef49cda

.travis.yml
```diff
@@ -2,10 +2,11 @@ language: ruby
...
   - 2.1.0
-  - 2.2.0
+  - 2.2.4
   - jruby-19mode
 
 env:
   - RAILS='~> 4.0.13'
   - RAILS='~> 4.1.10'
   - RAILS='~> 4.2.1'
+  - RAILS='~> 5.0.beta'
```
BenMorganIO, line 12: Might have to update this to `5.0.0.beta1`.

BenMorganIO:
Maybe drop support for 2.0 and 2.1 for this gem? I think that would be a good idea since rails is moving to the 2.2 area and this is largely a rails gem.

halostatue:
I’m fine with any direction on this. IMO, the way to fix the tests would be to switch to multiple Gemfiles rather than using an ENV setting, because I’m not sure that you can exclude from a Travis test matrix based on the ENV setting.

jhawthorn:
We can skip older versions of ruby with rails 5 using the exclude matrix feature of CircleCI. https://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix

Ideally I'd like to keep compatibility with older versions of ruby unless there's a compelling reason to drop it.

halostatue:
Cool. Apparently you can have the ENV be part of the exclusion. I will update the `travis.yml` to reflect this.

lstrzebinczyk:
Is there any chance to push this? Updating paranoia is blocking me from updating to 5.0.0, and issue seems to be with travis settings, not anything with the code.

halostatue:
Unless someone else wants to pick this up, I should be able to finalize the matrix for this on the weekend.

lstrzebinczyk:
@halostatue Any progress?

jhawthorn:
Thanks for the work! This was cherry-picked with modifications into master, which should now support rails 5.

seanhandley:
Nice one @jhawthorn

seanhandley:
Is a release coming soon?

PR #291, opened 2015-12-28
Don't try to add to null transactions

shosti:
Fixes https://github.com/rubysherpas/paranoia/issues/274

It looks like ActiveRecord refactored the Transaction class hierarchy in
a non-backwards-compatible
way (https://github.com/rails/rails/pull/16363), which breaks
`add_to_transaction` for null transactions (I think it probably
qualifies as a bug). The workaround is to not try to use
`add_to_transaction` in null transactions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,9 @@
 language: ruby
 rvm:
   - 2.0.0
-  - 2.1.0
-  - 2.2.0
+  - 2.1.8
+  - 2.2.4
+  - 2.3.0
   - jruby-19mode
 
 env:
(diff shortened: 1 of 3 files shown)


hail2skins:
Interested in seeing this added as I have the same issue.   Interestingly, I have acts_as_paranoid on one model and that works fine.   But on a model I just added it to, it fails, so not sure how that works exactly :).

krainboltgreene:
Please update the branch D:

krainboltgreene:
Also point to master?

shosti:
OK updated. Do you mean "core"?

krainboltgreene:
@shosti Ignore that comment, I was a tiny bit confused! :)

PR #297, opened 2016-01-13
Call `add_to_transaction` on #restore! as well

Empact:
Under certain conditions, combining the use of paranoia and
counter_culture can result in callbacks not being run. This is
perhaps due to an interaction between after_commit_action and
user-defined after_commit callbacks, but unfortunately I have been
unable to pin this down in tests.

Note this makes restore! roughly symmetric with delete/destroy in
its existing use of add_to_transaction.


Empact:
I made some fairly extensive efforts to repro under test then gave up. What do you guys think?

shosti:
BTW this will probably be affected by https://github.com/rubysherpas/paranoia/issues/274 (and should have a fix similar to https://github.com/rubysherpas/paranoia/pull/309).

PR #302, opened 2016-01-30
Rails 5 compatible branch

pigeonworks:
Update adding Rails5 branch. Passes all tests with travis.ci on Rails 5 compatible Ruby versions. Includes the fix #291 to resolve null transaction issue in ActiveRecord.


pigeonworks:
Don't want to merge this into the Rails4 branch, want to create a new branch.

pigeonworks:
Reopening as I cannot find a way for the pull request to create a new branch, but this is not intended to be merged into the Rails4 branch, but to create a new one.

jhawthorn:
Is there a compelling reason to make a new branch? It seems like there are minimal changes required between support of rails 4.2 and rails 5

pigeonworks:
No compelling reason if it will work successfully as a single branch. Rails 5 has different Ruby requirements so the travis setup was easier. I created it as a separate branch initially not knowing what changes would need to be made to support Rails 5, plus it follows the pattern of having separate branches for Rails 3 and Rails 4.

Archao:
:+1:

jhawthorn:
Thanks for the work. The master branch should now support rails 5.

PR #303, opened 2016-02-02
Rails 5 Support, Require ruby 2.0

jhawthorn:
Alternative to #284 #290 #302. Thanks to @pigeonworks @halostatue and @gagalago whose work this is based on. I tried to credit as best as possible using cherry-picks and the changelog.

This code will form the basis of paranoia `2.2.0`. This explicitly requires ruby >= 2.0, which we stopped testing against in `2.1.2`, but probably still worked. Because of the ruby 2.0 requirement we are now testing against jruby 9.0.5.0, which required a change because of inconsistent behaviour of `alias` vs `alias_method`. Currently jruby doesn't seem to work with rails 5.0.0.beta1.1, so that combination is an allowed failure on travis.

I am merging this into the `rails4` branch, which I am going to maintain as a branch compatible with both rails 4 and rails 5. I might rename the branch `master` to avoid confusion (and because I don't want to work on the rails3 branch other than bug fixes).


halostatue:
Thanks, @jhawthorn. This is excellent.

PR #305, opened 2016-02-04
Add scope without_deleted

rbr:
When using the option :without_default_scope it makes sense to have a scope to exclude paranoia-deleted records, which is essentially the paranoia_scope.
Aliased it to without_deleted as opposed to the existing with_deleted scope.

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -219,7 +219,8 @@ class ActiveRecord::Base
     def self.paranoia_scope
       where(paranoia_column => paranoia_sentinel_value)
     end
-    
+    class << self; alias_method :without_deleted, :paranoia_scope end
+
     unless options[:without_default_scope]
       default_scope { paranoia_scope }
     end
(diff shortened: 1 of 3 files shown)


lib/paranoia.rb
```diff
@@ -219,8 +219,10 @@ def self.acts_as_paranoid(options={})
     def self.paranoia_scope
       where(paranoia_column => paranoia_sentinel_value)
     end
-    
-    unless options[:without_default_scope]
+
+    if options[:without_default_scope]
```
jhawthorn: We might as well add it even when `without_default_scope` is false.

Maybe someone will find use from `Model.unscoped.without_deleted`

Other than that LGTM :+1:

rbr: I agree, makes sense. I made the changes and rebased.

krainboltgreene:
Thanks!

jhawthorn:
Awesome. Thanks rbr

rbr:
:+1:

PR #308, opened 2016-02-08
Detailing the changes for version 3

krainboltgreene:


paranoia.gemspec
```diff
@@ -4,21 +4,15 @@ require File.expand_path("../lib/paranoia/version", __FILE__)
 Gem::Specification.new do |s|
   s.name        = "paranoia"
   s.version     = Paranoia::VERSION
-  s.platform    = Gem::Platform::RUBY
-  s.authors     = ["radarlistener@gmail.com"]
-  s.email       = []
-  s.homepage    = "http://rubygems.org/gems/paranoia"
+  s.authors     = ["Kurtis Rainbolt-Greene (@krainboltgreene) <me@kurtisrainboltgreene>"]
+  s.homepage    = "https://rubysherpas.github.io/paranoia"
```
jhawthorn, line 8: This is a 404. Maybe should be `https://github.com/rubysherpas/paranoia`

README.md
```diff
@@ -20,10 +22,10 @@ For Rails 3, please use version 1 of Paranoia:
 gem "paranoia", "~> 1.0"
 ```
 
-For Rails 4 or 5, please use version 2 of Paranoia:
+For Rails 4 or 5, please use version 3 of Paranoia:
 
 ``` ruby
-gem "paranoia", "~> 2.0"
+gem "paranoia", "3.0.0"
```
jhawthorn, line 28: We should encourage semver with `"~> 3.0"`

jhawthorn:
@krainboltgreene I noticed there's a new default branch named `core`, which is a great idea to avoid confusion related to rails 4 vs rails 5 compatibility :+1:. Could I suggest we use `master` instead, which I think is more widely used and understood?

krainboltgreene:
@jhawthorn:

Yes, I actually plan on using semver metadata to denote releases for Rails 3, Rails 4, and Rails 5 support.

I prefer not to use the word master for the trunk branch. I think any small confusion is worth the inclusiveness.

jhawthorn:
@krainboltgreene 
I'm not sure what you mean by using semver that way? Why would we not use `"~> 3.0"`?

If `master` is out, how about `trunk`? `core` is not a helpful name. To me it implies the "core" component of a set of components. It doesn't make much sense as a branch name.

p:
> To me it implies the "core" component of a set of components. It doesn't make much sense as a branch name.

Agreed.

Which rails version(s) is the "core" branch going to support?

krainboltgreene:
Latest stable.

p:
Then you could call that branch "master" and the other branches "railsX" based on which rails versions they support. E.g. "master" and "rails3". Seems pretty clear to me which branch to use when.

krainboltgreene:
I'm not naming the branch master.

On Thu, Feb 11, 2016 at 8:25 PM, Oleg Pudeyev notifications@github.com
wrote:

> Then you could call that branch "master" and the other branches "railsX"
> based on which rails versions they support. E.g. "master" and "rails3".
> Seems pretty clear to me which branch to use when.
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/rubysherpas/paranoia/pull/308#issuecomment-183174594.

## 

Kurtis Rainbolt-Greene, Software Developer
Founder of Difference Engineers

2985 San Marino St.
Los Angeles, CA 90006
202-643-2263

mathieujobin:
does not look like there is anything worth keeping here, closing...

PR #333, opened 2016-06-12
Add options for uniqueness validations. Resolves #319

### LINKED ISSUE #319: How can I use uniqueness validation for all records, include deleted.
In my application I need to validate email uniqueness for all records, not only not deleted.

rbr:
For with_default_scope add a `paranoia: :with_deleted` option when you want to ensure uniqueness on all records.
For without_default_scope add a `paranoia: :without_deleted` option when you want to ensure uniqueness on only non-deleted records.

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,6 +181,30 @@ Client.restore(id, :recursive => true)
 client.restore(:recursive => true)
 ```
 
+If you want to validate uniqueness but want to include deleted records:
+
+``` ruby
+class Client < ActiveRecord::Base
+  acts_as_paranoid
+
+  validates :name, uniqueness: { paranoia: :with_deleted }
+
+  ...
+end
+```
+
+If you don't use the default_scope and want to validate uniqueness but want to exclude deleted records:
+
+``` ruby
+class Client < ActiveRecord::Base
+  acts_as_paranoid without_default_scope: true
+
+  validates :name, uniqueness: { paranoia: :without_deleted }
+
+  ...
+end
+```
+
 For more information, please look at the tests.
 
 #### About indexes:
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -212,8 +212,9 @@ class ActiveRecord::Base
     alias_method :destroy_without_paranoia, :destroy
 
     include Paranoia
-    class_attribute :paranoia_column, :paranoia_sentinel_value
+    class_attribute :paranoia_column, :paranoia_sentinel_value, :paranoid_by_default
 
+    self.paranoid_by_default = !options[:without_default_scope]
     self.paranoia_column = (options[:column] || :deleted_at).to_s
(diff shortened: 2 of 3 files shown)


rbr:
This will break behaviour for people already using the `without_default_scope` option together with a uniqueness validation. Currently it will exclude deleted records, after this merge it will include them, and you'd need to use the `paranoia: :without_deleted` option to get the same behaviour again .

Wasn't sure if we should handle that case as the `without_default_scope` option hasn't been put into an official version yet.

denys281:
@BenMorganIO  @jhawthorn  Could  we hope that it sometimes will be merged?

nghiabk:
Don't custom paranoid gem.You can write:
`validates :name, uniqueness: {conditions: ->{with_deleted}}`

README.md
```diff
@@ -181,6 +181,30 @@ Client.restore(id, :recursive => true)
 client.restore(:recursive => true)
 ```
 
+If you want to validate uniqueness but want to include deleted records:
+
+``` ruby
+class Client < ActiveRecord::Base
+  acts_as_paranoid
+
+  validates :name, uniqueness: { paranoia: :with_deleted }
```
nghiabk, line 190: Don't custom paranoid gem.You can write:
validates :name, uniqueness: {conditions: ->{with_deleted}}

README.md
```diff
@@ -181,6 +181,30 @@ Client.restore(id, :recursive => true)
...
+end
+```
+
+If you don't use the default_scope and want to validate uniqueness but want to exclude deleted records:
+
+``` ruby
+class Client < ActiveRecord::Base
+  acts_as_paranoid without_default_scope: true
+
+  validates :name, uniqueness: { paranoia: :without_deleted }
```
nghiabk, line 202: Don't custom paranoid gem.You can write:
validates :name, uniqueness: {conditions: ->{with_deleted}}

jerko-culina: > Don't custom paranoid gem.You can write:
> validates :name, uniqueness: {conditions: ->{with_deleted}}

Doesn't work for me

daneshd: Doesn't work for me either in `rails v4.2.10` using MySQL. Traced it a bit and I found that rails wasn't actually removing the scope via the `unscope` method because of the way the relation query is set up. By the time it hits [here](https://github.com/rails/rails/blob/v4.2.10/activerecord/lib/active_record/validations/uniqueness.rb#L29), the relation is wrapped and chained with an `AND` query. It doesn't get through [this case](https://github.com/rails/rails/blob/v4.2.10/activerecord/lib/active_record/relation/query_methods.rb#L912) statement to actually remove the column from the scope because the relation `rel` is a `Arel::Nodes::And` at that point.

I just ended up making my own validator which manually checked for uniqueness against the unscoped records to move forward.

PR #338, opened 2016-07-06
Support multiple sentinel values

yonekawa:
We want to use multiple sentinel values sometimes.
For example, we use status column.

``` rb
class StatusColumnModel < ActiveRecord::Base
  # { 0 => initialized, 1: active, 2: deleted }
  acts_as_paranoid column: :status, sentinel_value: [0, 1]

    { deleted_at: nil, status: 0 }
  end

    { deleted_at: current_time_from_proper_timezone, status: 2 }
  end
end
```

But paranoia does not support it because use `!=` operator.
We need to use `NOT IN` and `include?`.


kei2100:
+1

yonekawa:
Anyone?  How do you think?

PR #346, opened 2016-07-19
Fixes a problem of ambiguous table names when using only_deleted method and joining tables that have a scope on `with_deleted`

### LINKED ISSUE #26: Possible ambiguous column issue for only_deleted
Hello.

Suppose there were 2 database tables named alpha and beta and both had a deleted_at column...

One would get an ambiguous column "deleted_at" if the following were attempted:

```
Alpha.joins(:beta).only_deleted
```

Perhaps this:

```
def only_deleted
  unscoped {
    where("deleted_at is not null")
  }
end
```

Should be this:

```
def only_deleted
  unscoped {
    where("#{self.table_name}.deleted_at is not null")
  }
end
```

Or perhaps I am attempting to use this method in a manner that is not intended.

thromera:
Fixes #26 and #27 (that are still present). 

This bug would happen when overriding sentinel with `active` field

```
# paranoid.rb
module Paranoid
  extend ActiveSupport::Concern

  # See index section here: https://github.com/radar/paranoia

  included do
    acts_as_paranoid column: :active, sentinel_value: true
    include Overrides
  end

  # If not defined in a separate module, Paranoia will be prepended before Paranoid, causing these methods to be later
  # in the call chain
  module Overrides
      {
        deleted_at: nil,
        active: true
      }
    end

      {
        deleted_at: current_time_from_proper_timezone,
        active: nil
      }
    end
  end
end
```

```
#post.rb
class Post < ActiveRecord::Base
    include Paranoid

    belongs_to :author, -> {with_deleted}
    belongs_to :event, -> {with_deleted}

    validates_presence_of :author, :event
end
```

```
#author.rb
class Author < ActiveRecord::Base
    include Paranoid

    has_many :posts, -> {with_deleted}, dependent: :destroy
end
```

```
#event.rb
class Event < ActiveRecord::Base
    include Paranoid

    has_many :posts, -> {with_deleted}
end
```

```
Post.only_deleted.joins(:event)
2.2.3 :001 > Post.only_deleted.joins(:event)
  Post Load (10.1ms)  SELECT "posts".* FROM "posts" INNER JOIN "events" ON "events"."id" = "posts"."event_id" AND "events"."active" = ? WHERE ("active" IS NULL OR "active" != 't')  [["active", "t"]]
ActiveRecord::StatementInvalid: SQLite3::SQLException: ambiguous column name: active: SELECT "posts".* FROM "posts" INNER JOIN "events" ON "events"."id" = "posts"."event_id" AND "events"."active" = ? WHERE ("active" IS NULL OR "active" != 't')
```

### DIFF
--- a/test/paranoia_test.rb
+++ b/test/paranoia_test.rb
@@ -40,6 +40,8 @@ def setup!
     'unparanoid_unique_models' => 'name VARCHAR(32), paranoid_with_unparanoids_id INTEGER',
     'active_column_models' => 'deleted_at DATETIME, active BOOLEAN',
     'active_column_model_with_uniqueness_validations' => 'name VARCHAR(32), deleted_at DATETIME, active BOOLEAN',
+    'paranoid_model_with_belongs_to_active_column_model_with_has_many_relationships' => 'name VARCHAR(32), deleted_at DATETIME, active BOOLEAN, active_column_model_with_has_many_relationship_id INTEGER',
+    'active_column_model_with_has_many_relationships' => 'name VARCHAR(32), deleted_at DATETIME, active BOOLEAN', 
     'without_default_scope_models' => 'deleted_at DATETIME'
   }.each do |table_name, columns_as_sql_string|
     ActiveRecord::Base.connection.execute "CREATE TABLE #{table_name} (id INTEGER NOT NULL PRIMARY KEY, #{columns_as_sql_string})"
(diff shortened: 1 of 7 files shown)


test/paranoia_test.rb
```diff
@@ -5,6 +5,7 @@
 
 test_framework = defined?(MiniTest::Test) ? MiniTest::Test : MiniTest::Unit::TestCase
 
+
```
BenMorganIO: ?

test/paranoia_test.rb
```diff
@@ -1105,6 +1122,47 @@ def paranoia_destroy_attributes
...
+  def paranoia_destroy_attributes
+    {
+      deleted_at: current_time_from_proper_timezone,
+      active: nil
+    }
+  end
+end
+
+
+
```
BenMorganIO: ?

BenMorganIO:
The commit for this doesn't lead to a Github profile and isn't owned by anyone. I'm a bit skeptical security wise.

thromera:
I'm currently in vacation, I will fix that next week, my git config was probably wrong at that time.

thromera:
It should be fixed now. (I've also fixed the useless new lines.)

BenMorganIO:
Cool. I'd love to get this in, but it looks like you need to rebase the latest changes from the `core` branch.

thromera:
I'm closing this PR since I've messed up with my rebase. Sorry about that. 
Good news is https://github.com/rubysherpas/paranoia/pull/379 should be OK to merge.

PR #365, opened 2016-11-28
Compatability with Rails 5.1: build_relation accepts 3 parameters - wit…

iFurt:
…hout `table`

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -263,8 +263,8 @@ require 'paranoia/rspec' if defined? RSpec
 module ActiveRecord
   module Validations
     module UniquenessParanoiaValidator
-      def build_relation(klass, table, attribute, value)
-        relation = super(klass, table, attribute, value)
+      def build_relation(klass, *args)
+        relation = super(klass, *args)
         return relation unless klass.respond_to?(:paranoia_column)
         arel_paranoia_scope = klass.arel_table[klass.paranoia_column].eq(klass.paranoia_sentinel_value)
         if ActiveRecord::VERSION::STRING >= "5.0"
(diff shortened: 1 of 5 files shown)


lib/paranoia.rb
```diff
@@ -263,8 +263,9 @@ def paranoia_sentinel_value
 module ActiveRecord
   module Validations
     module UniquenessParanoiaValidator
-      def build_relation(klass, table, attribute, value)
-        relation = super(klass, table, attribute, value)
+      def build_relation(*args)
```
BenMorganIO: Just bike shedding 🚲 🏠, but you could do:

```ruby
def build_relation(klass, *args)
  relation = super(klass, *args)
  ...
end
```

This way you don't have to say `klass = args.first`.

BenMorganIO:
Classic JRuby failing...

BenMorganIO:
For 5.0.0.1, it still takes 4 arguments: https://github.com/rails/rails/blob/v5.0.0.1/activerecord/lib/active_record/validations/uniqueness.rb#L53

**But...** Rails 5.1 will not: https://github.com/rails/rails/blob/master/activerecord/lib/active_record/validations/uniqueness.rb#L52

Maybe update your title for Rails 5.1 compatibility?

iFurt:
Sure, thanks for pointing, Ben 👍

BenMorganIO:
Also, thank you for this!

BenMorganIO:
Thanks. Could you rebase and squash them both into one commit and update the message to read 5.1 as well?

iFurt:
Seems that I've messed up with rebase..

PR #367, opened 2016-11-29
Add explicit language about dependent: :destroy

supremebeing7:
My team and I had a lengthy conversation about this because there were different interpretations of what the README meant. This may be beating a dead horse a bit, but I think it's an important enough piece of this great gem that it merits perhaps a little bit extra verbosity to make sure it's well understood.


BenMorganIO:
I like it :+1:

PR #371, opened 2016-12-12
Use ActiveSupport.on_load to correctly re-open ActiveRecord::Base

iaankrynauw:
https://github.com/rubysherpas/paranoia/issues/335

I don't really know how to write a test for this. Although I have checked that it fixes the issue.

In Rails5 `Rails.application.config.active_record.belongs_to_required_by_default = true` gets added in `new_framework_defaults.rb`.

When opening `ActiveRecord::Base` this gets overridden and requires `belongs_to model, required: true`.

The fix waits for `ActiveSupport.on_load(:active_record)` to load before opening `ActiveRecord::Base`.

```
class ParanoidModel < ActiveRecord::Base
  belongs_to :parent_model
  acts_as_paranoid
end

expect(ParanoidModel.create).to_not be_valid
expect(ParanoidModel.create(parent_model_id: 1)).to be_valid
```

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -205,56 +205,58 @@ module Paranoia
   end
 end
 
-class ActiveRecord::Base
-  def self.acts_as_paranoid(options={})
-    alias_method :really_destroyed?, :destroyed?
-    alias_method :really_delete, :delete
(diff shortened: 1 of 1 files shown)


BenMorganIO:
Thanks @iaankrynauw :+1:

PR #372, opened 2016-12-15
Fix counter cache double decrement problem.

marshall-lee:
Sometimes paranoia could decrement counter cache twice even if record is already destroyed.

In our app we have invalid counter cache column values, internal investigation showed that it was due to race condition while destroying the same record (double-click on the 'delete' button).

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -74,30 +74,29 @@ module Paranoia
 
   def destroy
     transaction do
+      add_to_transaction
       run_callbacks(:destroy) do
-        result = delete
-        next result unless result && ActiveRecord::VERSION::STRING >= '4.2'
-        each_counter_cached_associations do |association|
-          foreign_key = association.reflection.foreign_key.to_sym
-          next if destroyed_by_association && destroyed_by_association.foreign_key.to_sym == foreign_key
-          next unless send(association.reflection.name)
-          association.decrement_counters
+        raise ActiveRecord::ReadOnlyRecord, "#{self.class} is marked as readonly" if readonly?
+        affected_records = paranoia_destroy_row
+        next self unless ActiveRecord::VERSION::STRING >= '4.2'
+        if affected_records > 0
+          each_counter_cached_associations do |association|
+            foreign_key = association.reflection.foreign_key.to_sym
+            next if destroyed_by_association && destroyed_by_association.foreign_key.to_sym == foreign_key
+            next unless send(association.reflection.name)
+            association.decrement_counters
+          end
         end
-        result
+        self
       end
     end
   end
 
   def delete
-    raise ActiveRecord::ReadOnlyRecord, "#{self.class} is marked as readonly" if readonly?
-    if persisted?
-      # if a transaction exists, add the record so that after_commit
-      # callbacks can be run
-      add_to_transaction
-      update_columns(paranoia_destroy_attributes)
-    elsif !frozen?
-      assign_attributes(paranoia_destroy_attributes)
-    end
+    # if a transaction exists, add the record so that after_commit
+    # callbacks can be run
+    add_to_transaction if persisted?
+    paranoia_destroy_row
     self
   end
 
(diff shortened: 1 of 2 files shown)


lib/paranoia.rb
```diff
@@ -74,30 +74,29 @@ def self.extended(klazz)
...
+      add_to_transaction
       run_callbacks(:destroy) do
-        result = delete
-        next result unless result && ActiveRecord::VERSION::STRING >= '4.2'
-        each_counter_cached_associations do |association|
-          foreign_key = association.reflection.foreign_key.to_sym
-          next if destroyed_by_association && destroyed_by_association.foreign_key.to_sym == foreign_key
-          next unless send(association.reflection.name)
-          association.decrement_counters
+        raise ActiveRecord::ReadOnlyRecord, "#{self.class} is marked as readonly" if readonly?
```
BenMorganIO, line 79: Pity we just can't use [`ActiveRecord::Persistence#_raise_readonly_record_error`](https://github.com/rails/rails/blob/a9d72f6e47d42ce2b9f7ad9f957ef1bd4bb6c800/activerecord/lib/active_record/persistence.rb#L594)

lib/paranoia.rb
```diff
@@ -74,30 +74,29 @@ def self.extended(klazz)
...
+        affected_records = paranoia_destroy_row
+        next self unless ActiveRecord::VERSION::STRING >= '4.2'
```
BenMorganIO, line 81: Looks like in the tests we could have a different case for this line here.

lib/paranoia.rb
```diff
@@ -74,30 +74,29 @@ def self.extended(klazz)
...
+        if affected_records > 0
+          each_counter_cached_associations do |association|
+            foreign_key = association.reflection.foreign_key.to_sym
+            next if destroyed_by_association && destroyed_by_association.foreign_key.to_sym == foreign_key
```
BenMorganIO, line 85: If only we were locked on Ruby 2.3 :trollface:

lib/paranoia.rb
```diff
@@ -203,6 +202,33 @@ def restore_associated_records
...
+                         .where(primary_key => self[primary_key])
+                         .update_all(attributes)
+
+      if affected_records > 0
+        attributes.each do |k, v|
+          raw_write_attribute(k, v)
+        end
+      end
+
+      affected_records
```
BenMorganIO, line 224: This is a bit high on my mental ABC. Maybe pull some of this out to private methods?

lib/paranoia.rb
```diff
@@ -203,6 +202,33 @@ def restore_associated_records
 
     clear_association_cache if destroyed_associations.present?
   end
+
+  def paranoia_destroy_row
+    attributes = paranoia_destroy_attributes
+
+    if persisted?
+      model = self.class
```
BenMorganIO, line 210: lol :+1:

test/paranoia_test.rb
```diff
@@ -889,6 +889,22 @@ def test_counter_cache_column_update_on_destroy#_and_restore_and_really_destroy
     assert_equal 0, parent_model_with_counter_cache_column.reload.related_models_count
   end
 
+  # TODO: it seems impossible to fix double decrement on Rails < 4.2
+  #       because in these versions decrement on destroy
+  #       is being performed in before_destroy callback :(
```
BenMorganIO, line 894: Maybe add something to the readme? This way users have more of a reason to invest in upgrading their rails applications.

lib/paranoia.rb
```diff
@@ -203,6 +202,33 @@ def restore_associated_records
 
     clear_association_cache if destroyed_associations.present?
   end
+
+  def paranoia_destroy_row
```
BenMorganIO, line 206: Are you able to add some Yard docs for this method? It'll make the review easier for me and for future users sifting through this code :)

lib/paranoia.rb
BenMorganIO: There should honestly be a `readonly!?` method that would raise for us...

lib/paranoia.rb
```diff
@@ -203,6 +202,33 @@ def restore_associated_records
...
+    attributes = paranoia_destroy_attributes
+
+    if persisted?
+      model = self.class
+      primary_key = model.primary_key
+      affected_records = model
+                         .unscoped
+                         .without_deleted
+                         .where(primary_key => self[primary_key])
+                         .update_all(attributes)
```
BenMorganIO, line 216: Hey, could you explain this portion? I love how neatly laid out this is.

marshall-lee:
Bump!

cc @BenMorganIO

BenMorganIO:
@marshall-lee I left comments but discovered Github had changed to add an opt-in to leave them. They were written a long time ago. Apologies. Now I have to go through every PR I've commented on and do that opt-in.

excid3:
Unfortunately, I don't believe this fixes the situation of double decrement when you were to do a `destroy` followed by a `really_destroy!`.

PR #311, opened 2016-02-21
Touch record on paranoia-destroy. Fixes #296

rbr:
Touch record on destroy by leveraging the paranoia_destroy_attributes.
Applied the same to the restore-method as this eliminates the extra query.

As pointed out by @ccleung in #296, touch on destroy was broken in #245 / b2b8d19 (verified by running the included test against 2.1.3 and 2.1.4). Since update_columns is used now (reasons are pointed out in that commit), update-timestamps were no longer set by ActiveRecord.

PR #377, opened 2017-01-28
Touch record on paranoia-destroy. Fixes #296

### LINKED ISSUE #296: Record not touched on soft destroy
I could have sworn this worked as expected (and by paranoia design) that a records `updated_at` attribute gets touched when calling `destroy`. However, thats not happening. I looked at the code, and see a `touch` method call under the `restore` method, but not the `destroy` method. Am I missing something?

rbr:
Touch record on destroy by leveraging the paranoia_destroy_attributes.
Applied the same to the restore-method as this eliminates the extra query.

As pointed out by @ccleung in #296, touch on destroy was broken in #245 / b2b8d19 (verified by running the included test against 2.1.3 and 2.1.4). Since update_columns is used now (reasons are pointed out in that commit), update-timestamps were no longer set by ActiveRecord.

(Recreation of PR #311, as PR-repo was gone so no new commits could be pushed.)

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -110,7 +110,6 @@ module Paranoia
         if (noop_if_frozen && !@attributes.frozen?) || !noop_if_frozen
           write_attribute paranoia_column, paranoia_sentinel_value
           update_columns(paranoia_restore_attributes)
-          touch
         end
         restore_associated_records if opts[:recursive]
       end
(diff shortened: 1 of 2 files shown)


basiszwo:
Any chance someone whose allowed to merge can do so? Would be awesome! 👍

BenMorganIO:
@basiszwo done!

PR #383, opened 2017-02-20
Add recovery window feature

### LINKED ISSUE #359: Only recover records that have been destroyed through dependent: :destroy callback?
Hi, thanks for a great gem!

I'm sorry if this has been discussed earlier, but I couldn't find anything about it. The feature that I miss the most in this library is to be able to recursively restore only objects that have been destroyed through the `depent: :destroy` callback.

The `acts_as_paranoid` gem seems to have this [feature](https://github.com/ActsAsParanoid/acts_as_paranoid#recovery) and I wonder if there are any reasons for `paranoia` not to have it?

I'd like to help to work on a solution for this if there is interest in it.

konto-andrzeja:
Resolves #359.

Adds option to restore only those dependently destroyed associated records that were deleted within a timespan - before or after the object upon which they depend. The `:recovery_window` param is added to `restore` method to define the timespan.

The default behaviour is not modified, so not defining `:recovery_window` restores everything.

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,6 +181,14 @@ Client.restore(id, :recursive => true)
 client.restore(:recursive => true)
 ```
 
+If you want to restore a record and only those dependently destroyed associated records that were deleted within 2 minutes of the object upon which they depend:
+
+``` ruby
+Client.restore(id, :recursive => true. :recovery_window => 2.minutes)
+# or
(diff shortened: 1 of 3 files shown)


BenMorganIO:
Thanks!

PR #384, opened 2017-02-23
Fix UniquenessParanoiaValidator on rails 5.1.0.beta1

jhawthorn:
The number of arguments to `build_relation` has changed. Fortunately, we only use the first argument and it hasn't changed.

I didn't add 5.1.0.beta1 to the `.travis.yml`, as there are bound to be major changes in the upcoming beta versions. But the specs do pass when running `RAILS=5.1.0.beta1 rake`

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -284,8 +284,8 @@ require 'paranoia/rspec' if defined? RSpec
 module ActiveRecord
   module Validations
     module UniquenessParanoiaValidator
-      def build_relation(klass, table, attribute, value)
-        relation = super(klass, table, attribute, value)
+      def build_relation(klass, *args)
+        relation = super
         return relation unless klass.respond_to?(:paranoia_column)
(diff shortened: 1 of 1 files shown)


BenMorganIO:
Merci @jhawthorn!

lib/paranoia.rb
```diff
@@ -284,8 +284,8 @@ def paranoia_sentinel_value
 module ActiveRecord
   module Validations
     module UniquenessParanoiaValidator
-      def build_relation(klass, table, attribute, value)
-        relation = super(klass, table, attribute, value)
+      def build_relation(klass, *args)
+        relation = super
```
BenMorganIO, line 288: Actually, when we call `super`, shouldn't we pass in `klass` and `args`?

frederikspang: As far as I know `super` passes exact same arguments, while `super()` would pass none. (And of course `super(klass)` would pass only class, etc.

PR #385, opened 2017-02-23
Replace assert_equal nil with assert_nil

jhawthorn:
Fixes all deprecation warnings in our tests

### DIFF
--- a/test/paranoia_test.rb
+++ b/test/paranoia_test.rb
@@ -117,9 +117,9 @@ class ParanoiaTest < test_framework
     model.remove_called_variables     # clear called callback flags
     model.destroy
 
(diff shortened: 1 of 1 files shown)


BenMorganIO:
Looks like a rebase is needed for this :)

edwardmp:
this can be closed as already fixed by #390

AlexVPopov:
@edwardmp Why do you think #390 has resolved this? I just cloned master and the warnings still appear. Besided #390 has not changed these assertions.

edwardmp:
@AlexVPopov I'm sorry I linked the wrong pull request. The right one that fixes this is #389, which has been merged into master.

Issue #380, opened 2017-02-15
Deleted resources can still be associated to other resources


edwardmp:
I'm not sure whether this is actually expected to happen as this use case may not be within the intended scope of paranoia.

I have a model (let's call it `Card`) that acts as paranoid. I have another model (let's call it 'Event') that doesn't act as paranoid but `belongs_to` a `Card`. If I have an instance of a `Card` that was soft destroyed, I can still assign it to a new `Event` when I create it.

Is this expected behavior when using this gem? If so, does anyone have any suggestions as to how to prevent this (e.g. adding a custom validator checking that associations are not soft deleted before creating a resource)?

BenMorganIO:
@edwardmp I would add a validation. I think this may be out of the scope of paranoia for creating new resources with the intention of using an object that has been marked as deleted. Some people use paranoia on products and change the `deleted_at` to `unavailable_at`, yet still want to perform operations on them.

Maybe:

```ruby
class Event < ApplicationRecord
  belongs_to :card

  validate :card_is_alive

  private

  def card_is_alive
    return if card.deleted_at.nil?
    errors.add(:card, 'has been deleted')
  end
end
```

edwardmp:
@BenMorganIO 
Thanks. Seems we are on the same page as I opted for this in the meantime:
```
class AssociationNotSoftDestroyedValidator < ActiveModel::EachValidator
  def validate_each(record, attribute, value)
    # if association is soft destroyed, add an error
    if value.present? && value.deleted?
      record.errors[attribute] << 'has been soft-deleted'
    end
  end
end
```

This can then be used as:
` validates :card, association_not_soft_destroyed: true`

This might be of use to someone else. Although I agree that by default this would be out of paranoia's scope, might it be an idea to include this validator with paranoia for people to use?

BenMorganIO:
@edwardmp if you'd like to make a PR to add this validation, I would love to get it merged in :)

vraravam:
I have been thinking of the same issue - but from a different angle. In my case, I would still like the UI and the model-level validations to go through if I edit the `Event` (taking the same associations as above for illustrating my example scenario) - and as long as the associated `Card` was not changed, I don't want to raise a validation exception. But, if the `card_id` is changed, then it should point to a value that is *not* soft-deleted. Thus, depending on the action being performed, I would like the UI and the validations to work seamlessly for
1. a create action; vs
2. an update action that was not trying to point to a soft-deleted `Card`; vs 
[...]

PR #387, opened 2017-02-25
#380 Introduced a helper method that can be configured in the AR mode…

vraravam:
…l so as to retrieve the associated parent records irrespective of whether the actual parent has been soft-deleted or not.

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -234,6 +234,8 @@ module Paranoia
   end
 end
 
+require 'paranoia_helper'
+
 ActiveSupport.on_load(:active_record) do
   class ActiveRecord::Base
     def self.acts_as_paranoid(options={})
@@ -244,6 +246,7 @@ ActiveSupport.on_load(:active_record) do
       alias_method :destroy_without_paranoia, :destroy
 
       include Paranoia
+      include ParanoiaHelper
       class_attribute :paranoia_column, :paranoia_sentinel_value
 
       self.paranoia_column = (options[:column] || :deleted_at).to_s
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -234,7 +234,7 @@ module Paranoia
   end
 end
 
-require 'paranoia_helper'
+require 'association_helper'
 
 ActiveSupport.on_load(:active_record) do
   class ActiveRecord::Base
@@ -246,7 +246,7 @@ ActiveSupport.on_load(:active_record) do
       alias_method :destroy_without_paranoia, :destroy
 
       include Paranoia
-      include ParanoiaHelper
+      include AssociationHelper
       class_attribute :paranoia_column, :paranoia_sentinel_value
 
       self.paranoia_column = (options[:column] || :deleted_at).to_s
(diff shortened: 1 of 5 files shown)


vraravam:
@BenMorganIO - any updates on this?

lib/paranoia.rb
```diff
@@ -232,6 +234,7 @@ def self.acts_as_paranoid(options={})
       alias_method :destroy_without_paranoia, :destroy
 
       include Paranoia
+      include ParanoiaHelper
```
BenMorganIO: Are we able to `extend ParanoiaHelper`? That way we don't need the fancy `ActiveSupport::Concern`.

lib/paranoia_helper.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+module ParanoiaHelper
```
BenMorganIO: I'd probably call this something else. `ParanoiaHelper` isn't very immediate as to its purpose.

lib/paranoia_helper.rb
```diff
@@ -0,0 +1,45 @@
...
+    #     collection_for_edit :sales_person, nil, name_prefix: :salesmen
+    #     collection_for_edit :sales_person, -> { User.salesmen }, name_prefix: :salesmen   # This is equivalent to the above
+    #   end
+    #
+    # The method will evaluate the scope_block if specified to get all the values for the edit dropdown. If the scope_block is nil,
+    # the method will try to deduce it from the association. Again, if there is a scope defined for the association class with the
+    # same name as option 'name_prefix' or the pluralized version of the associated symbol, then that will be used to fetch all
+    # records for the dropdown
+    #
+    def collection_for_edit(associated_sym, scope_block = nil, options = {})
```
BenMorganIO: There's a lot going on inside of this method. Given, this codebase is legacy, but with this helper we have a chance to make things more readable. Are we able to use some private methods and reduce the ABC here?

lib/paranoia.rb
vraravam: @BenMorganIO : I can try to implement this, but, given that this gem is anyway dependent on `rails >= 4.0`, would this make sense?

lib/paranoia_helper.rb
vraravam: @BenMorganIO - sure, i can break this down into multiple private methods.

lib/paranoia.rb
BenMorganIO: When I read this line here I thought it was adding instance methods, then I discovered it was doing something to add class methods. It's a simple change, uses way less code, and is more idiomatic over a concern to include class level methods without any macros.

vraravam:
@BenMorganIO - i have made a couple of small changes. I'm not sure how much more the ABC can be reduced - any suggestions?

vraravam:
@BenMorganIO - After a long wait, I have rebased from the upstream branch and would like to get this PR merged. If not, I will close the PR. Please suggest how to go forward.

vraravam:
@BenMorganIO - any decision please?

mathieujobin:
I'm sorry @vraravam we never got back to you after your last update... 
I think your changes actually look good. I don't know if anybody else wanted to use them. That's all.

mathieujobin:
Thanks again for your contribution

PR #388, opened 2017-03-01
Changelog: Add missing bang in `#really_destroy!`

patbl:
Also, it might be a good idea to change this to "acts_as_paranoid for Rails 3, 4 & 5."

screenshot 2017-03-01 14 28 04


BenMorganIO:
@patbl Unfortunately I don't have that much access to make that change.

The change also needs to be rebased for me to merge :)

patbl:
OK, I rebased it.

PR #389, opened 2017-03-07
Added association not soft destroyed validator

edwardmp:
See #380. 

While we're at it I also included some changes so the following warning doesn't show up anymore:
`Use assert_nil if expecting nil from [..]. This will fail in MT6.`

### DIFF
--- a/test/paranoia_test.rb
+++ b/test/paranoia_test.rb
@@ -21,6 +21,7 @@ def setup!
     'paranoid_model_with_foreign_key_belongs' => 'parent_model_id INTEGER, deleted_at DATETIME, has_one_foreign_key_id INTEGER',
     'paranoid_model_with_timestamps' => 'parent_model_id INTEGER, created_at DATETIME, updated_at DATETIME, deleted_at DATETIME',
     'not_paranoid_model_with_belongs' => 'parent_model_id INTEGER, paranoid_model_with_has_one_id INTEGER',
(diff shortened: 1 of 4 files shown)


BenMorganIO:
Why the merge commit in e3eeb53?

BenMorganIO:
Looks like 51f9541 occurred in #385, yet @jhawthorn hasn't rebased. So I'll take yours :)

edwardmp:
@BenMorganIO because I first edited the README directly through GitHub, and then fixed the test warnings but forgot to pull first ;)

test/paranoia_test.rb
```diff
@@ -1265,6 +1278,12 @@ class NotParanoidModelWithBelong < ActiveRecord::Base
   belongs_to :paranoid_model_with_has_one
 end
 
+class NotParanoidModelWithBelongsAndAssocationNotSoftDestroyedValidator < NotParanoidModelWithBelong
+    belongs_to :parent_model
+    validates :parent_model, association_not_soft_destroyed: true
+end
+
+
```
BenMorganIO: extra new line here

BenMorganIO:
@edwardmp awesome. Could you add this to the Changelog using the same format as in 2.2.1 and see if you can remove the merge commit? It's fine if you can't, they're tricky things merge commits.

Thanks for all your work on getting this guy in!!!

test/paranoia_test.rb
edwardmp: fixed

edwardmp:
@BenMorganIO all fixed!

BenMorganIO:
Thanks!

PR #393, opened 2017-03-25
Test with rails 5.1 and ruby 2.4.0

mfazekas:
Updated travis.yml with rails 5.1 and ruby 2.4

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,25 +2,25 @@ sudo: false
 language: ruby
 cache: bundler
 rvm:
-  - 2.1.10
   - 2.2.6
   - 2.3.3
+  - 2.4.1
   - jruby-9.1.6.0
 
 env:
   matrix:
-    - RAILS='~> 4.1.16'
     - RAILS='~> 4.2.7.1'
     - RAILS='~> 5.0.0.1'
+    - RAILS='~> 5.1.0.rc1'
 
 matrix:
   exclude:
-    - env: RAILS='~> 5.0.0.1'
-      rvm: 2.1.10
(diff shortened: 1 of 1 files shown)


BenMorganIO:
@mfazekas awesome! Are you able to remove Rails 4.1 actually? We drop support as new rails versions come out :). (Keeps the CI fast and makes adding new features or fixing bugs easier for everyone.)

mfazekas:
I've removed 4.1 from travis. (In case you actually break 4.1 you might want to increase major version because of SemVer)

.travis.yml
```diff
@@ -5,22 +5,27 @@ rvm:
   - 2.1.10
```
BenMorganIO: Man, I remember the day when I was super excited for Ruby 2.1, but nowadays I think we can even remove this from Travis as well. 2.2 is quite compatible and gives that restriction from the `?.` operator.

.travis.yml
```diff
@@ -5,22 +5,27 @@ rvm:
...
   - 2.2.6
   - 2.3.3
+  - 2.4.0
```
BenMorganIO: I don't want to keep asking you to keep going back and making changes to something that should be quite simple... But, it looks like the latest is 2.4.1 :p

BenMorganIO:
Merged. I just realized, your commit is a lie now :p lol

aried3r:
Why is Rails 5.1.0 not being tested with Ruby 2.2?

mfazekas:
@aried3r i guess it was an overlook

PR #407, opened 2017-05-24
Fix indentation on ActiveRecord::Base#I_AM_THE_DESTROYER

moskyb:
This PR fixes the inconsistent indentation in the `ActiveRecord::Base#I_AM_THE_DESTROYER` method. This was necessary as I use this method in production regularly.

Thanks to @samueljseay and @rpearce for pointing this out.

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -271,11 +271,11 @@ ActiveSupport.on_load(:active_record) do
     # Pretty please.
     def self.I_AM_THE_DESTROYER!
       # TODO: actually implement spelling error fixes
-    puts %Q{
-      Sharon: "There should be a method called I_AM_THE_DESTROYER!"
-      Ryan:   "What should this method do?"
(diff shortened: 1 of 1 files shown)


rpearce:
Nice work! On another note here, when the method is called, it'll include a newline, and all the text will be indented as much as the method is. Might I make a suggestion?
```rb
def self.I_AM_THE_DESTROYER!
  # TODO: actually implement spelling error fixes
  puts <<~CONVO
    Sharon: "There should be a method called I_AM_THE_DESTROYER!"
    Ryan:   "What should this method do?"
    Sharon: "It should fix all the spelling errors on the page!"
  CONVO
end
```
then you'll get
```
irb(main):028:0> SomeClass.I_AM_THE_DESTROYER!
Sharon: "There should be a method called I_AM_THE_DESTROYER!"
Ryan:   "What should this method do?"
Sharon: "It should fix all the spelling errors on the page!"
=> nil
```
Also, this is all a bit tongue-in-cheek.

PR #408, opened 2017-05-30
⚠ instance variable @_disable_counter_cache not initialized

amatsuda:
Here's a fix for a trivial Ruby-level warning.

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -174,7 +174,7 @@ module Paranoia
   private
 
   def each_counter_cached_associations
-    !(defined?(@_disable_counter_cache) && @_disable_counter_cache) && defined?(super) ? super : []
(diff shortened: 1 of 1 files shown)


lib/paranoia.rb
```diff
@@ -174,7 +174,7 @@ def really_destroy!
   private
 
   def each_counter_cached_associations
-    !@_disable_counter_cache && defined?(super) ? super : []
+    !(defined?(@_disable_counter_cache) && @_disable_counter_cache) && defined?(super) ? super : []
```
BenMorganIO: Merci! I think you can also delete the `defined?(super)` check here as well if you want. We've dropped support for Rails 4.1 and this was meant for that old version.

amatsuda: @BenMorganIO Done!

PR #412, opened 2017-07-13
Fixes `really_destroy!` behavior with `sentinel_value`

### LINKED ISSUE #326: really_destroy! updates column with datetime even when using sentinel_value
Using paranoia 2.1.5, I have a model setup using the `sentinel_value` option, as [documented in the README under "Indexes"](https://github.com/rubysherpas/paranoia#about-indexes), e.g:

``` ruby
add_column :clients, :active, :boolean
add_index :clients, [:group_id, :active], unique: true

class Client < ActiveRecord::Base
  # optionally have paranoia make use of your unique column, so that
  # your lookups will benefit from the unique index
  acts_as_paranoid column: :active, sentinel_value: true

    {
      deleted_at: nil,
      active: true
    }
  end

    {
      deleted_at: current_time_from_proper_timezone,
      active: nil
    }
  end
end
```

When calling `really_destroy!` on this model the following warning is generated by Rails (4.2.6):

```
DEPRECATION WARNING: You attempted to assign a value which is not explicitly `true` or `false` (2016-05-17 02:02:23 UTC) to a boolean column. Currently this value casts to `false`. This will change to match Ruby's semantics, and will cast to `true` in Rails 5. If you would like to maintain the current behavior, you should explicitly handle the values you would like cast to `false`.
```

This is because paranoia is writing the current datetime into the paranoia column (`active` in this example), at this line: https://github.com/rubysherpas/paranoia/blob/core/lib/paranoia.rb#L146. I think this should probably use `paranoia_destroy_attributes` instead?

However it's not clear to me why it is even updating the column if it then goes and deletes the row from the database? Why update the column at all?

steverice:
This removes the assumption from `really_destroy!` that there is
only one column value, the deletion date, that needs to be updated
on a model that is being `really_destroy!`ed. Instead, use the
defined `paranoia_destroy_attributes` as they are used in `destroy`
and other calls.

This prevents an issue where the DBMS will complain when trying to
`really_destroy!` one of these models, given that paranoia is
attempting to insert a DateTime value into a boolean (or otherly-
typed) column.

Fixes #326

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -165,7 +165,7 @@ module Paranoia
             association_data.really_destroy!
           end
         end
-        write_attribute(paranoia_column, current_time_from_proper_timezone)
+        update_columns(paranoia_destroy_attributes)
         destroy_without_paranoia
       end
     end
--- a/test/paranoia_test.rb
+++ b/test/paranoia_test.rb
@@ -546,6 +546,14 @@ class ParanoiaTest < test_framework
     assert model.instance_variable_get(:@real_destroy_callback_called)
   end
 
+  def test_really_destroy_behavior_for_active_column_model
+    model = ActiveColumnModel.new
+    model.save
+    model.really_destroy!
+
+    refute ParanoidModel.unscoped.exists?(model.id)
+  end
+
   def test_really_delete
     model = ParanoidModel.new
     model.save


jhawthorn:
Thank you! Merged as #425.

PR #416, opened 2017-08-08
Feature/dirty tracking

dgmora:
I took what the first commit from https://github.com/rubysherpas/paranoia/pull/375

As far as I understood, `update_columns` does not track changes with `Dirty` because it does not use the regular attribute setters, it uses `raw_write_attribute`. `raw_write_attribute` also clears the changes in 5.1.x. See [here](https://github.com/rails/rails/blob/v5.1.3/activerecord/lib/active_record/attribute_methods/dirty.rb#L86) (which is called from `update_columns`. `super` is in [persistence](https://github.com/rails/rails/blob/v5.1.3/activerecord/lib/active_record/persistence.rb#L325)).

Adding `deleted_at_will_change!` and `changes_applied` works except for rails 5.0.x, couldn't find the reason. It looks like the [AttributeMutationTracker](https://github.com/rails/rails/blob/v5.0.5/activerecord/lib/active_record/attribute_mutation_tracker.rb) does not track the change in `deleted_at` correctly

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -147,10 +147,10 @@ module Paranoia
   alias :deleted? :paranoia_destroyed?
 
   def paranoia_update_columns(attributes)
+    update_columns(attributes)
     attributes.keys.each do |key|
       send("#{key}_will_change!")
     end
-    update_columns(attributes)
     changes_applied
   end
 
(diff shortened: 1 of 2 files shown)


mathieujobin:
can you rebase or merge `core` into your branch ?

mathieujobin (commented):
👀

PR #423, opened 2017-11-02
Add paranoia_destroy and paranoia_delete aliases

jhawthorn:
This allows users to be explicit about wanting to do a paranoia soft-delete instead of a normal hard-delete.


tvdeyen:
Ah, too late :(

I think the names of these methods should have been `paranoid_*`. Would have made them more readable.

But this is fine as well....

PR #426, opened 2017-11-03
Use self.class.transaction to avoid potential naming conflicts

brendancarney:
Although maybe not a good idea, some models will define a method `transaction`. 

This change uses the `transaction` method on `ActiveRecord::Base` instead.

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -74,7 +74,7 @@ module Paranoia
   end
 
   def destroy
-    transaction do
+    ActiveRecord::Base.transaction do
       run_callbacks(:destroy) do
         @_disable_counter_cache = deleted?
         result = delete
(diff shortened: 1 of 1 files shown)


harryhlai (approved):

jhawthorn:
Thanks for the suggestion, but we should not do this.

Different records may be from different databases, so it's important to call transaction on the correct model.

There should be no expectation that replacing methods defined by ActiveRecord will work. ActiveRecord even itself be calling `transaction`, on the model.

brendancarney:
Good point. 

What about the class method instead?
`self.class.transaction`

Like the `restore!` method.

brendancarney:
I pushed that change to this branch.  This makes it consistent across the three methods that use the transaction block.

PR #428, opened 2017-11-03
Fixed has_one bug with really_destroy! after destroy

harryhlai:
When an object and its associated objects have been soft deleted, the associations return empty collections and nil because of the default scoping. The logic for restore correctly handles this, but the logic for really_destroy (for an object that has already been soft deleted) fails for has_one associations. To see an example of this failing, look at the test case added in this PR; that test case fails with the current core branch.

This PR fixes that by refactoring the logic used in restore, and re-using it for really_destroy.


mathieujobin:
I am only seeing this now... I resolved the conflict simply by keeping both sides.
@harryhlai do you think it still makes sense to merge this?
can you update and see why the tests are failing?

PR #429, opened 2017-11-07
Ensure deleted_at, updated_at share the same value on delete

why-el:
Previously, on a delete call, `current_time_from_proper_timezone` was called
twice, which yields two different values, leading Rails to save two distinct timestamps in 
the DB, if one compares at a sub-millisecond level. 

This leads to problems when decisions need to be made as to whether a record was updated
after it was deleted (whatever the reason for this might be). Under current arrangements, the response
will always be yes if the system prefers a higher precision, even if that is not the case. 

The solution is to memoize the value and share it between the two columns on save. This saves the values to an acceptable precision as dictated by Rails and the OS platform, but in any case it would cease to be a Paranoia problem.

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -170,13 +170,17 @@ module Paranoia
   end
 
   def paranoia_destroy_attributes
+    timezoned_current_time = current_time_from_proper_timezone
+
     {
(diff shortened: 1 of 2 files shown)


dannysmith:
👍

PR #430, opened 2017-11-16
Improve method checks for cached associations

hkraji:
There is an issue that I've encountered where there is no `super` for method `each_counter_cached_associations` for given class. This happens when doing `restore(recursive: true)` on corresponding object. 

Rails version:

```ruby
gem 'rails', '4.1.16'
```

Not sure what is the root cause of it, but this is easy fix for it.

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -160,7 +160,7 @@ module Paranoia
   private
 
   def each_counter_cached_associations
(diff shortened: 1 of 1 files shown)


jhawthorn:
Thanks for the PR. Is it possible to add a test which exposes this issue?

PR #435, opened 2018-03-21
Monkeypatch activerecord relations to work with rails >=5.2.0rc2

bbonislawski:
Since this thing https://github.com/rails/rails/commit/b0fc04aa3af338d5a90608bf37248668d59fc881 got merged into rails, AR raises an exception and don't destroy record since relations seem not destroyed(because they're soft deleted instead of destroyed). 

We can monkeypatch ActiveRecord this way, but if someone has better solution, I will be glad.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ env:
     - RAILS='~> 4.2.0'
     - RAILS='~> 5.0.0'
     - RAILS='~> 5.1.0'
-    - RAILS='~> 5.2.0.beta2'
+    - RAILS='~> 5.2.0.rc2'
 
 matrix:
   allow_failures:
@@ -24,5 +24,5 @@ matrix:
       rvm: jruby-9.1.6.0
     - env: RAILS='~> 5.1.0'
       rvm: jruby-9.1.6.0
-    - env: RAILS='~> 5.2.0.beta2'
+    - env: RAILS='~> 5.2.0.rc2'
       rvm: jruby-9.1.6.0
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -1,4 +1,5 @@
 require 'active_record' unless defined? ActiveRecord
+require 'paranoia/active_record_patches'
 
 module Paranoia
   @@default_sentinel_value = nil
--- a/lib/paranoia/active_record_patches.rb
+++ b/lib/paranoia/active_record_patches.rb
@@ -1,0 +1,41 @@
+module HandleParanoiaDestroyedInBelongsToAssociation
+  def handle_dependency
+    return unless load_target
+
+    case options[:dependent]
+    when :destroy
+      target.destroy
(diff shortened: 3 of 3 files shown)


BenMorganIO:
@bbonislawski This is some great code, thanks!

Just one small issue is the `.class_eval`. This one is a little tricky in its behavior vs `.include` with a module. Are you able to place this nice code in a module then just do:

```ruby
module CatchHasOneDeletionException
  def delete(...)
    ...
  end
end

ActiveRecord::Associations::HasOneAssociation.include CatchHasOneDeletionException
```

Or which ever module name you feel better describes what you're achieving.

bbonislawski:
@BenMorganIO fixed. I had to use `prepend` instead of include and decided to move it to another file

BenMorganIO:
Looks like JRuby is failing: `NoMethodError: undefined method 'new_column'`

bbonislawski:
@BenMorganIO there is a problem with sqlite3 jdbc adapter(it's based on activerecord `~> 5.1.0` which doesn't match `5.2.0.rc2`) so I don't know if there is any sensible way to fix it before release of rails `5.2.0`  .

Any ideas?

BenMorganIO:
Crap. I'm not that into the ruby world ATM to be able to provide meaningful thoughts. Maybe someone else will have some thoughts on this. cc @jhawthorn

Edit: was going to ping him about this, but his profile on the solidus slack says he's vacationing until March 26th. I apologize @bbonislawski you may have to wait for a little for this one.

I think JRuby should remain to have support, but if they aren't going to get things working nicely for the upcoming rails 5.2, then we may have to sink them if they don't swim. I would recommend using your branch for the time being until a new release is cut.

.travis.yml
```diff
@@ -14,7 +14,7 @@ env:
     - RAILS='~> 4.2.0'
     - RAILS='~> 5.0.0'
     - RAILS='~> 5.1.0'
-    - RAILS='~> 5.2.0.beta2'
+    - RAILS='~> 5.2.0.rc2'
```
eitoball, line 17: @bbonislawski 

Don't you need to update `env: RAILS='~> 5.2.0.beta2'` line in `allowed_failures` section?

bbonislawski: Nice catch @eitoball . Fixed

mfazekas:
FWIW here is a gist to reproduce the issue:
https://gist.github.com/mfazekas/1a8778dfea45961ec1333506ca12d837

For now we're using the following hack in our soft deleteable models to work around:

```ruby
  def destroyed?
    if @in_save
      super
    else
      paranoia_destroyed?
    end
  end

  def save(*args)
    @in_save = true
    begin
      return super(*args)
    ensure
      @in_save = false
    end
  end
```

rubiii:
Thanks for this patch @bbonislawski 
I can confirm, that it fixes the problem for us.

Issue #285, opened 2015-12-08
Restoring a record does not take into account ActiveRecord validations


spint:
Since version 2.1.4, (uniqueness) validations checks only non-deleted records.
However, restoring a record that violates a Rails uniqueness constraint passes without an error.
Shouldn't the validations be considered before saving the restored record to the database?

andrewferk:
Looks to be the same as #319 and #333 is a proposed fix.

For now, you can enforce the uniqueness at the database by [updating your indexes](https://github.com/rubysherpas/paranoia#about-indexes). When you try to restore and it violates the database index, a mysql error should be raised.

PR #443, opened 2018-05-23
Fix incorrect default scope on uniqueness validation when without_default_scope is used

### LINKED ISSUE #319: How can I use uniqueness validation for all records, include deleted.

westonganger:
Hit this problem today. So I fixed it.

Fixes #319 & #285. Removes the need for PR #333


westonganger:
Just following up. This is a fairly important fix. Can we get this merged?

LakshyaKhatri:
@sevenseacat any update on this patch?

PR #448, opened 2018-06-27
Update README docs for paranoia_destroy_attributes

jmondo:
Since active is a boolean column, it would be better to send `false` than `nil`


jmondo:
Never mind, I get it now ;)

PR #449, opened 2018-07-04
Relax rails dependency and test run tests on edge rails version

dsalahutdinov:
Hi!

I'm gonna to test my Rails application with the master branch of Rails, so I suppose to relax Rails dependency what makes easier to maintain gem in the future 😄


dsalahutdinov:
Oops, seems something is broken...
I am going to fix it later

iggant:
have fixed all here: https://github.com/rubysherpas/paranoia/pull/456

PR #456, opened 2018-10-24
Allow activerecord 6.0

iggant:
We are trying to test our application on rails 6.0.0.alpha but we can't install this because of the version lock
Probably this will be helpful that we can try and report any issues with edge rails

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ rvm:
   - 2.3.5
   - 2.4.3
   - 2.5.0
+  - 2.6.1
   - jruby-9.1.6.0
 
 env:
@@ -33,4 +34,6 @@ matrix:
     - rvm: 2.2
       env: RAILS='master'
     - rvm: 2.3.5
+      env: RAILS='master'
+    - rvm: 2.4.3
       env: RAILS='master'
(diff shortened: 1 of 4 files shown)


jarednorman:
@jhawthorn The test suite for `solidus_core` passes with this branch of paranoia, so it's probably safe to merge and cut a release.

jhawthorn:
@jarednorman I don't want to remove the upper dependency altogether. This gem is guaranteed to break on some future rails version. But I would be happy to bump it to < 6.1

jarednorman:
Sure, I can PR a change to do that. 👍

iggant:
@jhawthorn done, also since rails 6.0 now support only ruby 2.5 exclude all below, 
Also add ruby 2.6.1 (escape 2.6.0 since it has a bug in http), of course this don't do anything with current gem but anyway

jarednorman:
Awesome, this should be good to go then. 🤘

matheusbn (approved):

iggant:
@matheusbn  resolved conflict so it can be merged now

jarednorman:
@jhawthorn I think this one is good, except for the CI issues, which are unrelated.

PR #462, opened 2019-02-11
Support ActiveRecord 6

jarednorman:
This pulls in @iggant's change from #456, but adds back in the upper version constraint on activerecord.


jarednorman:
#456 has this change, but without my silly typo, so closing this.

PR #467, opened 2019-03-08
.really_destroy! should not destroy has_many :through records

### LINKED ISSUE #466: .really_destroy! should not destroy has_many :through records
Despite of counter intuitive, Rails API states that `has_many :through` records must not be destroyed, but their associations https://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html#method-i-has_many

> If using with the `:through` option, the association on the join model must be a `belongs_to`, and the records which get deleted are the join records, rather than the associated records.

`paranoia` should have the same behavior

atd:
Despite of counter intuitive, Rails API states that `has_many :through` records must not be destroyed, but their associations https://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html#method-i-has_many

> If using with the `:through` option, the association on the join model must be a `belongs_to`, and the records which get deleted are the join records, rather than the associated records.

`paranoia` should have the same behavior 

Fixes #466

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -141,7 +141,8 @@ module Paranoia
       run_callbacks(:real_destroy) do
         @_disable_counter_cache = paranoia_destroyed?
         dependent_reflections = self.class.reflections.select do |name, reflection|
-          reflection.options[:dependent] == :destroy
+          reflection.options[:dependent] == :destroy &&
+            reflection.options[:through].blank?
         end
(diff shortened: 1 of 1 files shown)


mathieujobin (changes requested):
I see that current tests pass https://github.com/rubysherpas/paranoia/actions/runs/2026327334

But I would like to see a test case added for this and few more people advocating for this to be merged before taking side.

PR #475, opened 2019-07-09
Fix a call to nil on objects to destroy

Shigawire:
When a model has no :dependent option in their association, this is nil. `target.send(options[:dependent])` will fail in this case.


mathieujobin (changes requested):
Can you please add a test case to reproduce the problem this fixes ?

thanks

mathieujobin:
revisiting this... I really would like to see a test that reproduce the problem without the patch... 

Thanks for your contribution.

PR #476, opened 2019-10-10
Fix syntax to restore record that were deleted within 2 minutes

sachinjain007:


jarednorman:
This looks like a good change, but why is there a second empty commit with the same description as the one that actually contains the change? Mind removing that?

sachinjain007:
> This looks like a good change, but why is there a second empty commit with the same description as the one that actually contains the change? Mind removing that?

@jarednorman Thanks to pointing this. Last empty commit is rollback to 414541c which contains the actual change.

PR #482, opened 2019-11-28
Fix after_commit for Rails 6

hashwin:
Trigger after_commit on destroy to work with rails 6

This addresses https://github.com/rubysherpas/paranoia/issues/478

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -69,7 +69,7 @@ module Paranoia
           next unless send(association.reflection.name)
           association.decrement_counters
         end
-        @_trigger_update_callback = true
+        @_trigger_destroy_callback = true
         @_disable_counter_cache = false
         result
       end
@@ -80,6 +80,10 @@ module Paranoia
   def paranoia_destroy!
     paranoia_destroy ||
       raise(ActiveRecord::RecordNotDestroyed.new("Failed to destroy the record", self))
+  end
+
+  def trigger_transactional_callbacks?
+    super || @_trigger_destroy_callback && paranoia_destroyed?
   end
 
   def paranoia_delete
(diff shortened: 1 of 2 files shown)


williamweckl:
Can this be merged?

williamweckl:
Still waiting for it... Is the only thing that prevents me to updating to Rails 6. :(

qnighy:
@jhawthorn @sevenseacat Can one of the maintainers look at this? Thanks in advance!

zeitnot:
@jhawthorn @sevenseacat +1

jasonswett:
FWIW:

For whatever reason, when I upgraded to Paranoia 2.4.2, this change didn't come along with it. But when I put `gem 'paranoia', git: 'https://github.com/rubysherpas/paranoia', ref: '00a70579898f371975db907ea8e2ba3e8d737650'` in my Gemfile, the change did come through, and it fixed the bug.

sevenseacat:
There's been no release since this patch was merged - I don't have access to do that on Rubygems, unfortunately.

williamweckl:
@sevenseacat who has this access to make this happen?

sevenseacat:
@jhawthorn can you make a release with these changes?

smridge:
This change was included in a patch yet it is actually `BREAKING` if you are on Rails 5. If I'm putting it together correctly, when you update to Rails 6, the gem version this was included in needs to be updated at the same time?

mathieujobin:
hmm, according to the test suite, all tests pass with 5.2 and 5.1
https://github.com/rubysherpas/paranoia/actions/runs/2146476944

if there is a bug, we don't know about it.
unless someone provide a PR with a fix, I'd be more moving toward removing support for 5.x
and leaving a note as the best version to use with 5.x if any problem is encountered.

PR #483, opened 2019-12-02
Update JRuby version to 9.2.8.0 + remove EOL Ruby 2.2

donv:
`gem update --system` fails with Ruby 2.2.

This should also make #482 succeed.


donv:
Finally green!  Ready to merge?

donv:
Can this be merged?

PR #492, opened 2020-05-06
Warn if acts_as_paranoid is called more than once on the same model

### LINKED ISSUE #489: double acts_as_paranoid by mistake
Hi, this is more like a question not a real issue, it's because of a bug that was driving me crazy.

We had some resources that were raising some foreign key exceptions:

```PG::ForeignKeyViolation: ERROR:  update or delete on table "a" violates foreign key constraint "fk_rails_f207f40ee2" on table "b"```

For some reason, calling `really_destroy!` was not destroying associations it was just updating it's `deleted_at` column but this was only happening on one association (all of the associations have dependent :destroy), after a lot of debugging and testing I realized that in a bad merge (probably) we had a duplicate call to `acts_as_paranoid` so we had.

```
Class A < ActiveRecord::Base
  acts_as_paranoid
  acts_as_paranoid
end
```

I then took one of those calls out, and then it all started working properly.

I opened up the source code to try to understand what the heck was going on, and why having that double call to `acts_as_paranoid` makes the association to be soft deleted instead of really destroyed.

So that's why I'm here, just trying to understand what's happening and in case somebody else has an issue like this.

Rails 4.2.11.1
Paranoia 2.4.1

ignatiusreza:
which will breaks the method aliasing of `#destroy` and makes them lost
access to the original one defined by ActiveRecord

fix #489

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # paranoia Changelog
 
+## 2.5.2
+
+* [#492](https://github.com/rubysherpas/paranoia/pull/492) Warn if acts_as_paranoid is called more than once on the same model
+
+ [Ignatius Reza](https://github.com/ignatiusreza)
+
 ## 2.5.1
 
 * [#481](https://github.com/rubysherpas/paranoia/pull/481) Replaces hard coded `deleted_at` with `paranoia_column`.
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -242,6 +242,12 @@ end
 ActiveSupport.on_load(:active_record) do
   class ActiveRecord::Base
     def self.acts_as_paranoid(options={})
+      if included_modules.include?(Paranoia)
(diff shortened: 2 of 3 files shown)


mathieujobin (approved):

mathieujobin (approved):
This seems like a good idea... I wonder if my huge platform would be bitten by this...
I'd be temped to switch the `puts` for a raise if `Rails.env.test?` 
anyone else have an opinion on this?

I could include it in an upcoming release.
but needs version bump, changelog updated and rebased with `core` for tests to run.

ignatiusreza:
I rebased and added changelog

Given that paranoia is designed to be usable with activerecord outside of rails, I'm not quite sure how to best check for the rails environment. Any advice?

mathieujobin:
to answer your last question @ignatiusreza 

I recommend doing `if defined?(Rails)`

PR #501, opened 2020-11-23
Rails 6.1 Compatibility

walski:
Makes the gem usable in Rails 6.1.RC1 projects.

There is really not much too it besides a relaxation of the Rails version in the gemspec :shrug:

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -316,7 +316,7 @@ module ActiveRecord
       def validate_each(record, attribute, value)
         # if association is soft destroyed, add an error
         if value.present? && value.paranoia_destroyed?
-          record.errors[attribute] << 'has been soft-deleted'
+          record.errors.add(attribute, 'has been soft-deleted')
         end
       end
     end
--- a/paranoia.gemspec
+++ b/paranoia.gemspec
@@ -24,7 +24,7 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = '>= 2.0'
 
-  s.add_dependency 'activerecord', '>= 4.0', '< 6.1'
+  s.add_dependency 'activerecord', '>= 4.0', '< 6.2'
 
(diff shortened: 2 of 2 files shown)


walski:
We're using this in production now. Seems to *just work* ™

afomera:
We're not using this direct fork/branch in production (with the change to the Errors) but we're using the Rails 6.1rc2 gem and the current published gem and we've been stable. So I came here to bump the Gemspec for the expected official release and then saw this. Seems good to me!

joergschiller:
We also forked this repo and relaxed the gemspec to test our upgrade to Rails 6.1.

I think it would be a good idea to test paranoia with Rails 6.1 as well. So I added 6.1 to the build matrix:
https://github.com/rubysherpas/paranoia/compare/core...zinsbaustein:rails6.1

Maybe this is a good idea to incorporate into your pull request here?

stuckerab:
@joergschiller that sounds like a good idea!

I'm also waiting for this to get bumped up... before testing in a current project.

walski:
I like the idea but it seems very uncertain how much value Travis will add to OS projects like this one in the future (see e.g. https://news.ycombinator.com/item?id=25338983).

On the flip side there seems to be no downside to adjusting the build matrix for now 🤷 I'm happy to close this PR in favor of a new one from @joergschiller

stuckerab:
I had not heard that. Maybe switch to something like Semiphore? I think I
saw that they supported OSS projects with free minutes

anthonator:
What about GitHub Actions?

joergschiller:
@walski I opened a new PR https://github.com/rubysherpas/paranoia/pull/503 that also adds Rails 6.1 to the build matrix.

I know about the discussions regarding Travis but I think this is something that is out of the scope of this PR. It was building my own fork and it looks like Travis is [still building](https://travis-ci.com/github/rubysherpas/paranoia/builds/208223817) the new PR.

Thanks for your help so far!

PR #508, opened 2021-02-08
Support dependent: :destroy

### LINKED ISSUE #507: Association is not restored when using dependent: :destroy_async
Rails 6.1 introduces the new option `dependent: :destroy_async` (see rails/rails#40157). When using Paranoia, the associations are soft deleted in the background job as they should be but are not restored properly.

pomartel:
Fixes #507

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -193,7 +193,7 @@ module Paranoia
   # we called #destroy
   def restore_associated_records(recovery_window_range = nil)
     destroyed_associations = self.class.reflect_on_all_associations.select do |association|
-      association.options[:dependent] == :destroy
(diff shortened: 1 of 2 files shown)


pomartel:
This is more complicated than I first thought. Counter caches and permanent deletes also need to be handled correctly. I started implementing a custom `DestroyAssociationAsyncJob` but now I see that the gem maintainers don't accept new features so I will close this PR for now but I would be interested in digging deeper if there is a chance this feature can be accepted.

mathieujobin:
@pomartel I am definitely interesting seeing `destroy_async` support in paranoia

pomartel:
Sorry @mathieujobin, I opted to destroy records asynchronously in my own code instead (using a background job).

mathieujobin:
No problem

It sounds like a good solution

PR #513, opened 2021-09-17
Fix create callback called on destroy

bkDJ:
We were running into a weird bug in one of our projects while migrating from Rails 5.2 to Rails 6.0. It seems to be due to Rails using `#with_transaction_returning_status` when persisting models which takes care of setting the recently-added ivar `@_new_record_before_last_commit`. Paranoia overrides ActiveRecord's `#destroy` with `#paranoia_destroy` which unfortunately only uses `#transaction`, so when checking for which callbacks should be run, it was seeing a stale value of `@_new_record_before_last_commit`, therefore causing `after_create_commit` hooks to be called again.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -18,7 +18,7 @@ platforms :rbx do
   gem 'rubysl-test-unit'
 end
 
-rails = ENV['RAILS'] || '~> 5.2.0'
+rails = ENV['RAILS'] || '~> 6.0.4'
 
 if rails == 'master'
   gem 'rails', github: 'rails/rails'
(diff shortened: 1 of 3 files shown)


mathieujobin:
can you please rebase your branch with `core` so the tests run?
Thanks

rchasman:
Would really appreciate getting this fix into the wild :)

arlando:
@mathieujobin is the requirement just rebasing the branch to core and this will be GTG? thnx

mathieujobin:
I will have another review after I see all tests passing
 But it looked fine from what I remember.

mathieujobin (approved):
I just would like to see the tests running on CI ...

mathieujobin:
I have rebased this branch into a new branch. all tests passed. including with the test in #522 

I will release later 

https://github.com/rubysherpas/paranoia/actions/runs/2026193146

PR #515, opened 2021-10-04
Use GitHub Action instead of travis-ci.org

willnet:
CI doesn't work against recent PRs because building on travis-ci.org is ceased.

We can use travis-ci.com instead of travis-ci.org, but it has limitations to build time. So I think using GitHub Actions is the best CI to build this repository.

I migrated the settings while maintaining the original state as much as possible. Improvements to the configuration, such as adding Rails 7.0, will be done in the next PR.


willnet:
You can check the test results of this PR here.

[Use GitHub Action instead of travis-ci.org · willnet/paranoia@5689e21](https://github.com/willnet/paranoia/actions/runs/1302894789)

.github/workflows/test.yml
```diff
@@ -0,0 +1,55 @@
...
+
+on:
+  push:
+    branches: [ core ]
+  pull_request:
+    branches: [ core ]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
```
mathieujobin: 20.04 sounds a better bet for something aimed at servers

.github/workflows/test.yml
```diff
@@ -0,0 +1,55 @@
...
+    strategy:
+      matrix:
+        ruby-version:
+          - 2.3.8
+          - 2.4.5
+          - 2.5.3
+          - 2.6.3
```
mathieujobin, line 18: shall we take the opportunity to remove 2.3, and 2.4, add 2.7 and 3.0 ?
separate PR is better?

.github/workflows/test.yml
```diff
@@ -0,0 +1,55 @@
...
+          - rails_6_1
+        exclude:
+          - ruby-version: '2.3.8'
+            gemfile: rails_6_0
+          - ruby-version: '2.3.8'
+            gemfile: rails_6_1
+          - ruby-version: '2.4.5'
+            gemfile: rails_6_0
+          - ruby-version: '2.4.5'
+            gemfile: rails_6_1
```
mathieujobin, line 34: I think that was neat how they made it work with ENV vars for travis. it should be possible to keep the same pattern with github actions, at least that's what I am trying

mathieujobin:
:+1: on switching to Github actions. Run build is here https://github.com/willnet/paranoia/actions/runs/1302894789

All test pass ! 

shall we remove the `.travis.yml` file ?

mathieujobin:
Alternative https://github.com/rubysherpas/paranoia/pull/516

.github/workflows/test.yml
willnet: good point! 
changed at https://github.com/rubysherpas/paranoia/pull/515/commits/11908cdc36fbe35a0ce21888e20f4fab21e3dd57

radar:
Thanks, merged in #516 + #519.

PR #516, opened 2021-11-04
allow rails 7.0

mathieujobin:
Attempting an upgrade on Rails 7.
I need to bump this dependency.

PR #519, opened 2021-12-18
Rails7

radar:
Re-creation of #516, as the actions had not run on that PR.


radar:
If we're testing 5.1 + 5.2 + 6.1, we should also test 6.0.

radar:
Speaking of Rails versions: I'd be okay with this project dropping support for any Rails version prior to 5.1.

mathieujobin:
@radar I have tagged 2.5.0 and 2.5.1
did you want to push them to rubygems? or you wanted me to do it ? I don't have access at the moment.
I would need to share with you my private rubygem email for you to send the invite

PR #522, opened 2022-01-05
[DON'T MERGE] Add failing tests for association with abort on destroy

tricknotes:
paranoia destroys associations even if the destruction is aborted.
My expectation is what associations are not destroyed as the same as ActiveRecord's default.

@radar 
How do you think about this behavior?

----

I added reproduction code as failing tests.
Here is the result of this test cases.
``` console
$ RAILS='~> 6.1' bundle exec rake
Run options: --seed 21553

# Running:

...............F

Failure:
ParanoiaTest#test_destroy_behavior_for_association_with_abort [test/paranoia_test.rb:146]:
Expected: 0
  Actual: 1


rails test test/paranoia_test.rb:135

....................................F

Failure:
ParanoiaTest#test_destroy_bang_behavior_for_association_with_abort [test/paranoia_test.rb:161]:
Expected: 0
  Actual: 1


rails test test/paranoia_test.rb:149

..............................

Finished in 0.542657s, 152.9511 runs/s, 475.4384 assertions/s.
83 runs, 258 assertions, 2 failures, 0 errors, 0 skips
```

### DIFF
--- a/test/paranoia_test.rb
+++ b/test/paranoia_test.rb
@@ -31,6 +31,7 @@ def setup!
     'plain_models' => 'deleted_at DATETIME',
     'callback_models' => 'deleted_at DATETIME',
     'fail_callback_models' => 'deleted_at DATETIME',
+    'association_with_abort_models' => 'deleted_at DATETIME',
     'related_models' => 'parent_model_id INTEGER, parent_model_with_counter_cache_column_id INTEGER, deleted_at DATETIME',
     'asplode_models' => 'parent_model_id INTEGER, deleted_at DATETIME',
     'employers' => 'name VARCHAR(32), deleted_at DATETIME',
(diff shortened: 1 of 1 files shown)


tricknotes:
I don't have an idea to fix this tests yet.
However, I know using `prepend: true` avoids that.
``` diff
diff --git a/test/paranoia_test.rb b/test/paranoia_test.rb
index 81f80c6..482abdd 100644
--- a/test/paranoia_test.rb
+++ b/test/paranoia_test.rb
@@ -1169,7 +1169,7 @@ end
 class AssociationWithAbortModel < ActiveRecord::Base
   acts_as_paranoid
   has_many :related_models, class_name: 'RelatedModel', foreign_key: :parent_model_id, dependent: :destroy
-  before_destroy { |_|
+  before_destroy(prepend: true) { |_|
     if ActiveRecord::VERSION::MAJOR < 5
       false
     else
```

mathieujobin:
is this related to #513 ?

mathieujobin (approved):

PR #531, opened 2022-03-22
added a test case to cover the bug

sherif-nedap:


sherif-nedap:
Issue is occurring because of  this way of determining the deletion_time is not taking into account custom columns with datatype different than datetime 
in this case a Boolean `:active`

so here it will try to compare the `recovery_window_range` to the `active` column
and it should instead compare it to the `deleted_at`

this used to work fine before because `deleted_at` were hard coded
```
  def within_recovery_window?(recovery_window_range)
    return true unless recovery_window_range
    recovery_window_range.cover?(deletion_time)
  end
```

mathieujobin:
@sherif-nedap @sas1ni69 

I ran this in a debugger, I don't quite understand, why is the value of `active` being returned as `deletion_time` ?

That seems wrong... 
I can make the test pass by making this change... but I am not sure it makes sense

image

sas1ni69:
@mathieujobin yeah that doesn't seem right. I've separated both values from each other. I have a PR open here https://github.com/rubysherpas/paranoia/pull/532/files

Let me know if this works for you 🙏🏻

sas1ni69:
Also, how would I include @sherif-nedap 's test as his own commit?

Really appreciate the test case

PR #535, opened 2022-10-18
skip updating paranoia_destroy_attributes for records while really_destroy!

kortirso:
There is issue with destroying records when, for example:
- record had nil attribute, or optional belongs_to (database didn't have not null constaint)
- record was soft-deleted
- some validation was added, or belongs_to was changed to required state
- calling `record.really_destroy!` will fail during `update_columns`, because record can't be saved with some nil values

and there is some additional profit, `really_destroy!` will work faster without updating timestamps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # paranoia Changelog
 
+## 2.6.1
+
+* [#535](https://github.com/rubysherpas/paranoia/pull/535) Allow to skip updating paranoia_destroy_attributes for records while really_destroy!
+  [Anton Bogdanov](https://github.com/kortirso)
+
 ## 2.6.0
 
 * [#512](https://github.com/rubysherpas/paranoia/pull/512) Quote table names; Mysql 8 has keywords that might match table names which cause an exception.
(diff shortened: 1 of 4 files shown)


mathieujobin:
if I understand your patch properly, its opt-out by default, so we need to specify `really_destroy!(update_destroy_attributes: false)` explicitly, otherwise, it makes no difference, am I reading this right?

Also, can you update the Changelog and bump the version, please?

kortirso:
> if I understand your patch properly, its opt-out by default, so we need to specify `really_destroy!(update_destroy_attributes: false)` explicitly, otherwise, it makes no difference, am I reading this right?
> 
> Also, can you update the Changelog and bump the version, please?

yes, by default everything will work as usually, and only who wants to skip updating in the middle of really_destroy - they will call `really_destroy!(update_destroy_attributes: false)`

kortirso:
> if I understand your patch properly, its opt-out by default, so we need to specify `really_destroy!(update_destroy_attributes: false)` explicitly, otherwise, it makes no difference, am I reading this right?
> 
> Also, can you update the Changelog and bump the version, please?

changelog and readme are updated

kortirso:
@mathieujobin is it mergeable?

mathieujobin (approved):

mathieujobin:
Thanks for the ping, looks good

mathieujobin:
released

kortirso:
thank you

PR #541, opened 2023-06-05
Use correct time class for earlier Rails versions

emilong:
ActiveRecord pre-6.1 returned a `Time` object instead of a `DateTime` object, so this change updates the tests to use the correct class for the Rails version being tested.


mathieujobin (approved):
Weird that it all passed in November... Thanks

PR #546, opened 2023-08-24
Remove unused local variable and move local variables into conditional branch

ken3ypa:
Found dead code, minor enhancement point and fixed them：
- Removed unused local variable `association_class_name` introduced in commit 9937512664d2338ffe1fe3cd522e84dfbf0cafa6 .
- And relocated local variables referenced solely when `association_class.paranoid?` is true.
- Streamlined local variable assignments, aligning with the pattern used for `association_foreign_key`.

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -219,12 +219,9 @@ module Paranoia
         association_class = association.klass
         if association_class.paranoid?
           association_foreign_key = association.options[:through].present? ? association.klass.primary_key : association.foreign_key
-          association_find_conditions = if association.type
-                                          association_polymorphic_type = association.type
-                                          { association_polymorphic_type => self.class.name.to_s, association_foreign_key => self.id }
-                                        else
-                                          { association_foreign_key => self.id }
-                                        end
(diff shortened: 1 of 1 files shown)


lib/paranoia.rb
```diff
@@ -216,23 +216,15 @@ def restore_associated_records(recovery_window_range = nil)
...
-
         association_class = association.klass
         if association_class.paranoid?
+          association_foreign_key = association.options[:through].present? ? association.klass.primary_key : association.foreign_key
+          association_find_conditions = if association.type
+                                          association_polymorphic_type = association.type
+                                          { association_polymorphic_type => self.class.name.to_s, association_foreign_key => self.id }
+                                        else
+                                          { association_foreign_key => self.id }
+                                        end
```
mathieujobin: how about ....
```ruby
association_find_conditions = { association_foreign_key => self.id }
association_find_conditions[association.type] = self.class.name if association.type
```
appears clearer than the 6 lines above..

mathieujobin (commented):
Thanks for the optimization, just one comment

PR #548, opened 2023-10-05
Rails 7.1 Released 🚀

indyarocks:
Rails 7.1 released 🚀

https://github.com/rails/rails/releases/tag/v7.1.0

Trying to bump this dependency

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,11 +15,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        rails: ["~> 7.0.0", "~> 6.1.0", "~> 6.0.0"]
+        rails: ["~> 7.1.0", "~> 7.0.0", "~> 6.1.0", "~> 6.0.0"]
         ruby: ["3.2.2", "3.1.4", "3.0.6", "2.7.8"]
         include:
           - ruby: 3.2
             rails: 'edge'
+          - ruby: 3.2
+            rails: '~> 7.1.0'
           # single test failure with jruby
           #- ruby: jruby-9.4
           #  rails: '~> 7.0.0'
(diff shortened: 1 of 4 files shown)


frederikspang (approved):

paranoia.gemspec
```diff
@@ -24,7 +24,7 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = '>= 2.5'
 
-  s.add_dependency 'activerecord', '>= 5.1', '< 7.1'
+  s.add_dependency 'activerecord', '>= 5.1', '<= 7.1'
```
lsylvester: I think that this should be

```suggestion
  s.add_dependency 'activerecord', '>= 5.1', '< 7.2'
```

so that patch versions of 7.1 are also permitted.

.github/workflows/build.yml
```diff
@@ -15,11 +15,13 @@ jobs:
...
       fail-fast: false
       matrix:
-        rails: ["~> 7.0.0", "~> 6.1.0", "~> 6.0.0"]
+        rails: ["~> 7.1.0", "~> 7.0.0", "~> 6.1.0", "~> 6.0.0"]
         ruby: ["3.2.2", "3.1.4", "3.0.6", "2.7.8"]
         include:
           - ruby: 3.2
             rails: 'edge'
+          - ruby: 3.2
+            rails: '~> 7.1.0'
```
mathieujobin, line 24: That's duplicate run from the above matrix setting.

mathieujobin (commented):
Can you also bump the version and update the Changelog.

Please. 🙏🏽

.github/workflows/build.yml
indyarocks: This might not be redundant, as it ensures that you are testing against the latest 3.2.x version of Ruby.

mathieujobin: I don't think we need to test both 3.2.2 and the future 3.2.3, I rather update to 3.2.3 for every rails version when it comes out

mathieujobin (approved):

PR #551, opened 2024-01-29
fix: restore has_one with scope

zygzagZ:
When recursively restoring a has_one relation, Paranoia tries to restore any matching deleted record. It ignores the association scope. This can cause both false positives and false negatives when restoring.

This PR imitates [this commit](https://github.com/ActsAsParanoid/acts_as_paranoid/commit/d6a898e35789089706a8130f706720504a527662) from acts_as_paranoid. I am not sure if intentionally, but that commit introduced scope merging for has_one relations there.

> when looking for parent.alpha (or beta) to restore, loading association fails, because there are no undeleted records
> lib/paranoia.rb:204 `association_data = send(association.name)` = nil
> the procedure at lib/paranoia.rb:224 ignores association scope and restores the lowest-id record, :gamma
> effect: beta was never restored and is nil, gamma was restored incorrectly

I am not sure if `reflection.scope` is the correct way to recover association's scope, acts_as_paranoid does this differently. It works in AR 5.1-7.1 though.
I have tested this PR in ruby 3.2.2 in rails 7.1.3 and 6.1.7.6. It also does work in 5.1.7 application, but I failed to resolve gems in ruby 2.5.8/2.6.9 (and run the suite)

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -193,6 +193,16 @@ module Paranoia
     timestamp_attributes_for_update_in_model.each_with_object({}) { |attr,hash| hash[attr] = current_time_from_proper_timezone }
   end
 
+  def paranoia_find_has_one_target(association)
+    association_foreign_key = association.options[:through].present? ? association.klass.primary_key : association.foreign_key
+    association_find_conditions = { association_foreign_key => self.id }
+    association_find_conditions[association.type] = self.class.name if association.type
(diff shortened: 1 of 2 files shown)


zygzagZ:
I am not sure what happens with has_many relations, they might need patching too.

mathieujobin:
Thank you very much for this

PR #554, opened 2024-03-08
Support prebuilt counter cache association list

### LINKED ISSUE #552: Rails main - no superclass method `each_counter_cached_associations'
[Rails main deleted `each_counter_cached_associations` ](https://github.com/rails/rails/pull/49866/files#diff-0441501ea93f2580c66bc58d845fcc2ad9dc5110cce249ce184c0b59483e253bL211-L215)(private api). 
The next release of Rails will break paranoia. 

I am getting this error when targeting rails main: 
`NoMethodError: super: no superclass method `each_counter_cached_associations'`

JoeDupuis:
Fixes https://github.com/rubysherpas/paranoia/issues/552

Rails 7.2 will remove `each_counter_cached_associations` which was the hook to bypass incrementing/decrementing the counter. 
This change overrides the new counter cache list and uses it as the new hook to bypass updating the counter when it is present, else it falls back to the old behavior.

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -173,8 +173,25 @@ module Paranoia
 
   private
 
+  def counter_cache_disabled?
+    defined?(@_disable_counter_cache) && @_disable_counter_cache
+  end
+
+  def counter_cached_association_names
+    return [] if counter_cache_disabled?
+    super
+  end
+
   def each_counter_cached_associations
-    !(defined?(@_disable_counter_cache) && @_disable_counter_cache) ? super : []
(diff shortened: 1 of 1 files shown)


mathieujobin (approved):

mathieujobin:
Thanks a lot for this fix

rickychilcott:
Thank you for your fix @JoeDupuis! Given that Rails 7.2 is getting closer to being released, would you be willing to cut a new gem version @mathieujobin with this fix?

mathieujobin:
Thanks for the reminder @rickychilcott 

I will release that shortly

PR #558, opened 2024-08-07
Do not use sqlite3 v2.0.0 or later to fix CI

yoheimuta:
This PR updates the Gemfile to sqlite3 `~> 1.4` to pass the CI.

### Background

We have failed builds on GitHub Actions because sqlite3 version v2.0.0 was just released, and the current ActiveRecord doesn't support sqlite3 v2.

- ref. 
   - https://rubygems.org/gems/sqlite3/versions
   - https://github.com/rails/rails/pull/51592

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,7 +5,10 @@ sqlite = ENV['SQLITE_VERSION']
 if sqlite
   gem 'sqlite3', sqlite, platforms: [:ruby]
 else
-  gem 'sqlite3', platforms: [:ruby]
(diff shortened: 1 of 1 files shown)


mathieujobin:
I believe CI on the main branch is passing now

Are you sure this is still necessary??

yoheimuta:
No. After I created this PR, https://github.com/rubysherpas/paranoia/pull/563 fixed the issue.
Thank you for checking this PR. I'm closing it.

PR #559, opened 2024-08-08
feat: Trigger an after_commit callback when restoring a record

### LINKED ISSUE #457: After Restore callback executing before commit
Hi! 
I'm seeing that the after restore callback is getting run before the update commit.

`gem 'paranoia', '~> 2.4'`

```
class Flag < ApplicationRecord
  acts_as_paranoid

  after_restore :my_callback

  def my_callback
    p 'hi'
  end
end 

```

```
[7] pry(main)> flag.restore!
   (0.3ms)  BEGIN
  SQL (0.5ms)  UPDATE "flags" SET "deleted_at" = NULL, "updated_at" = '2018-10-28 12:55:34.948615' WHERE "flags"."id" = $1  [["id", 15]]
'hi'
   (5.9ms)  COMMIT
```

In my real callback I'm enqueuing a job that when executes it looks for the record in the db again, and if the job starts fast enough the state of the record is not yet updated.
Is this the expected behaviour for this callback? How can I get an after restore commit callback?

Thank you!

yoheimuta:
Fixes #457

This PR adds a new option called `after_restore_commit` so that Paranoia triggers a callback after committing to restore a record. The default value is `false`, so this change won't affect existing users.

Currently, there is no way to trigger a callback **after** commit. [The suggested workaround](https://github.com/rubysherpas/paranoia/issues/457#issuecomment-488716328) doesn't work since the `restore` method doesn't trigger `after_commit, on: :update` either.

📝 This branch is checked out from #558 to pass the CI tests.


yoheimuta:
@radar @mathieujobin Hi, I just wanted to see, is something wrong with this PR?

mathieujobin:
@yoheimuta thanks for the reminder. I will review.

For now, can you update or rebase and fix the merge conflict please?

yoheimuta:
@mathieujobin Thank you for the reply.
I just fixed the merge conflict.

mathieujobin (dismissed):
Looks like a great feature with no downside.

All tests passed except for rails edge (8.0) and it's unlikely related 

So I have no objections merging this 

I will wait a bit for others to chime in and release later this month

radar:
Looks good to me as well. Happy for you to handle the release when you're ready @mathieujobin.

mathieujobin (approved):

yoheimuta:
Thank you!

PR #563, opened 2024-08-12
Add support for Rails 7.2

andrewhampton:
This adds support for Rails 7.2. 

Everything is pretty straightforward with the exception of the required version of `sqlite3`. I see the `SQLITE_VERSION` env var can be used to specify a specific version of sqlite3, so maybe I should use that for testing rails 7.2?

closes #562

PR #572, opened 2024-12-05
Add support for Rails 8.0

chenillen:
- add support to 8.0
- add a test all file for 6.1 - 8.0 version


.ruby-version
```diff
@@ -0,0 +1 @@
+3.2.0
```
mathieujobin, line 1: not sure why we need this file, especially with `.0` ...

lib/paranoia.rb
```diff
@@ -1,7 +1,7 @@
 require 'active_record' unless defined? ActiveRecord
 
 if [ActiveRecord::VERSION::MAJOR, ActiveRecord::VERSION::MINOR] == [5, 2] ||
-   ActiveRecord::VERSION::MAJOR > 5
+   ActiveRecord::VERSION::MAJOR >= 6
   require 'paranoia/active_record_5_2'
```
mathieujobin, line 5: looking at this change, this looks badly named... do we need this patch for rails 6, 7 and 8 ?

Gemfile
```diff
@@ -20,7 +21,7 @@ if RUBY_ENGINE == 'rbx'
   end
 end
 
-rails = ENV['RAILS'] || '~> 6.0.4'
+rails = ENV['RAILS'] || '~> 6.1.0'
```
mathieujobin, line 24: we could move up the `rails` initialization, before `sqlite` and use the var instead, but I can see how its equivalent. so maybe its OK as-is

mathieujobin (commented):
These looks like good changes... I only have minor comments... but the title is miss-leading. this isn't adding support for Rails 8. it's only fixing small things and adding a script. Support for 8 is already merged in #563

.ruby-version
chenillen: Rails 8.0 requires at least 3.2.0 I think?

paranoia.gemspec
```diff
@@ -22,9 +22,10 @@ Gem::Specification.new do |s|
 
   s.required_rubygems_version = ">= 1.3.6"
 
-  s.required_ruby_version = '>= 2.7'
+  s.required_ruby_version = '>= 2.7.0'  # Base requirement for Rails 6.1
+  s.required_ruby_version = '>= 3.2.0' if ENV['RAILS']&.start_with?('~> 8')  # Rails 8.0 requirement
```
mathieujobin, line 26: +1

mathieujobin (changes requested):
its missing the update to the workflow file, so that CI runs for the new rails/ruby combinaison

.ruby-version
mathieujobin: that file isn't needed

Gemfile
mathieujobin: you should move up that line above `sqlite` so you can use it, allowing the default value.

Gemfile
```diff
@@ -5,7 +5,8 @@ sqlite = ENV['SQLITE_VERSION']
 if sqlite
   gem 'sqlite3', sqlite, platforms: [:ruby]
 else
-  gem 'sqlite3', '~> 1.4', platforms: [:ruby]
+  # Rails 8.0 requires sqlite3 2.x
+  gem 'sqlite3', ENV['RAILS']&.start_with?('~> 8') ? '~> 2.1' : '~> 1.4', platforms: [:ruby]
```
mathieujobin, line 9: I wonder, if `2.1+` works with rails 7 or 6, then we should turn the condition around and only use `~> 1.4` for older versions that don't support 2.x

ghost:
Hello,

> Support for 8 is already merged in https://github.com/rubysherpas/paranoia/pull/563

We're actually updating our codebase from 7.1 to 8.
And so far in the process, the bump from 7.2 to 8 is causing some trouble.
Anything I can do to help, because it appears it's not supporting Rails 8 yet 😅

mathieujobin:
thanks for your contribution.

PR #574, opened 2025-02-06
Fix really_destroy used against a has_one relationship that is soft deleted

reedstonefood:
I found a bug. The test explains the scenario but I'll explain it here as well.

Given:
```ruby
class MyModel < ActiveRecord::Base
  acts_as_paranoid
  has_one: linked_model, dependant: destroy
end

class LinkedModel < ActiveRecord::Base
  acts_as_paranoid
end

my_model = MyModel.create!(linked_model: LinkedModel.new)
```

- ✅ If you run `my_model.destroy` then both objects get soft-deleted as you would expect
- ❌ However if you then subsequently run `my_model.really_destroy` it does not destroy the associated `linked_model`.

This is because paranoia tries to access the linked model by running `self.send(name)`. This command will not pick up soft-deleted `has_one` associations (though it works fine for `has_many` associations) so some extra code is required in this scenario.

Note the reload in the test is required, as otherwise the version of `my_model` in memory still links to the `LinkedModel`.

### DIFF
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -182,6 +182,9 @@ module Paranoia
             association_data = self.send(name)
             # has_one association can return nil
             # .paranoid? will work for both instances and classes
+            if association_data.nil? && reflection.has_one? && reflection.klass.paranoid?
+              association_data = reflection.klass.only_deleted.find_by(reflection.foreign_key => self.id)
+            end
             next unless association_data && association_data.paranoid?
             if reflection.collection?
               next association_data.with_deleted.find_each { |record|
--- a/lib/paranoia.rb
+++ b/lib/paranoia.rb
@@ -183,7 +183,9 @@ module Paranoia
             # has_one association can return nil
             # .paranoid? will work for both instances and classes
             if association_data.nil? && reflection.has_one? && reflection.klass.paranoid?
-              association_data = reflection.klass.only_deleted.find_by(reflection.foreign_key => self.id)
+              query = { reflection.foreign_key => self.id }
+              query[reflection.type] = self.class.name if reflection.options[:as] # Handle polymorphic associations
+              association_data = reflection.klass.only_deleted.find_by(query)
             end
             next unless association_data && association_data.paranoid?
             if reflection.collection?
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # paranoia Changelog
 
+## 3.1.1 - July 14, 2026
+
+- [#574](https://github.com/rubysherpas/paranoia/issues/574) Fix really_destroy used against a has_one relationship that is soft deleted
+
 ## 3.1.0 - November 7, 2025
 
 - [#580](https://github.com/rubysherpas/paranoia/issues/580) Support Rails 8.0 and 8.1
--- a/lib/paranoia/version.rb
+++ b/lib/paranoia/version.rb
@@ -1,3 +1,3 @@
 module Paranoia
-  VERSION = '3.1.0'.freeze
+  VERSION = '3.1.1'.freeze
 end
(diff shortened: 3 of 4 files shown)


reedstonefood:
@mathieujobin any chance you could have a look at this please?

reedstonefood:
FYI the test failures on rails versions older than 7.1 are nothing to do with the changes in this MR, they are a wider issue that is detailed here: https://stackoverflow.com/questions/79360526/uninitialized-constant-activesupportloggerthreadsafelevellogger-nameerror

reedstonefood:
I've just spotted that this looks like a duplicate of #428, but as I'm actively engaged about this MR and it has working tests, it probably makes sense to progress with this one.

lib/paranoia.rb
```diff
@@ -182,6 +182,9 @@ def really_destroy!(update_destroy_attributes: true)
             association_data = self.send(name)
             # has_one association can return nil
             # .paranoid? will work for both instances and classes
+            if association_data.nil? && reflection.has_one? && reflection.klass.paranoid?
+              association_data = reflection.klass.only_deleted.find_by(reflection.foreign_key => self.id)
+            end
```
mathieujobin, line 189: @yoheimuta Can you chime in about this change?

mathieujobin:
is it possible to fix the tests for older version in this PR or in a different PR ?

lib/paranoia.rb
yoheimuta: @mathieujobin Looks good to me! 👍  

[nits] For `has_one` associations that are polymorphic (`foreign_type` column exists), `find_by` should also filter by `foreign_type` to avoid fetching unrelated records.  

Suggested improvement:  
```ruby
query = { reflection.foreign_key => self.id }
query[reflection.type] = self.class.name if reflection.options[:as] # Handle polymorphic associations
association_data = reflection.klass.only_deleted.find_by(query)
```
This ensures we correctly match records when dealing with polymorphic `has_one` associations.

mathieujobin: @reedstonefood Can you review this suggestion, see if that matches with your use case, etc.

thanks

reedstonefood:
> is it possible to fix the tests for older version in this PR or in a different PR ?

I've raised https://github.com/rubysherpas/paranoia/pull/576 for this.

lib/paranoia.rb
reedstonefood: This change works with the test suite here & in the product I'm working on, so LGTM.

I've also managed to add another test that fails without this change in place 👍 .

reedstonefood:
@mathieujobin please could this get re-reviewed?

mathieujobin:
Alright, I mean... I am not currently able to give a LGTM or if anything is wrong or can cause any problem to anyone...

I sure see that all tests passes 👍 

I can approve, merge and release in a few days... 

Can you bump the release number and update the change log?

Do you think it's worth increasing the minor or only the smaller revision is fine?

Other than that, I don't have capacity

reedstonefood:
@mathieujobin sorry for the long delay, I've bumped the version and added a changelog entry as requested.

This bug was first identified 9 years ago in https://github.com/rubysherpas/paranoia/pull/428, it would be good to fix it (and be able to remove the monkey patch in the codebase I'm working on 🙂).

---

> Do you think it's worth increasing the minor or only the smaller revision is fine?

The readme says:

> If you wish to actually destroy an object you may call really_destroy!. WARNING: This will also really destroy all dependent: :destroy records

which, is currently not true in this scenario. So, I think it's a patch only as it's making the gem function according to its published specification.

reedstonefood:
@mathieujobin please?

reedstonefood:
@mathieujobin @yoheimuta @mohammednasser-32 @radar @andrewhampton

Sorry for the mass ping, you are the most recent committers to this repository and am looking for someone who can move this forward. It's been reviewed, and I believe it's ready to merge.

I'd really appreciate it if someone could move this forward & fix this 9-year old bug for good.

PR #577, opened 2025-10-22
Loosen activerecord dependency to support Rails 8.1

dennispaagman:
Also added ruby 3.4 in the test matrix.

### DIFF
--- a/paranoia.gemspec
+++ b/paranoia.gemspec
@@ -24,7 +24,7 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = '>= 2.7'
 
-  s.add_dependency 'activerecord', '>= 6', '< 8.1'
+  s.add_dependency 'activerecord', '>= 6', '< 8.2'
 
   s.add_development_dependency "bundler", ">= 1.0.0"
   s.add_development_dependency "rake"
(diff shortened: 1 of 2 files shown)


davidwessman (approved):

msxavi:
Thank you for stepping up @dennispaagman! Who can approve the git workflows?

paranoia.gemspec
```diff
@@ -24,7 +24,7 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = '>= 2.7'
 
-  s.add_dependency 'activerecord', '>= 6', '< 8.1'
+  s.add_dependency 'activerecord', '>= 6', '< 8.2'
```
JulienDefrance, line 27: Should we instead have this be open-ended, to avoid similar complications whenever Rails 8.2 or 8.3 comes out?

dennispaagman: yeah I agree, just thought to make the smallest change possible first. I think having rails 8.x supported seems fine, the Active Record API is super stable.

mathieujobin:
Thank you

davidwessman:
@dennispaagman I guess this one can be closed now, since it has been resolved on main 🙂

PR #579, opened 2025-10-23
compatibility with rails 8.1

jules-w2:


msxavi:
Thanks for jumping onto this @jules-w2! FYI https://github.com/rubysherpas/paranoia/pull/577

binarygit:
When will this get merged? This is a blocker from upgrading

rmm5t (commented):
Needs additions to the build matrix: `.github/workflows/build.yml`

https://github.com/rubysherpas/paranoia/blob/core/.github/workflows/build.yml#L18

paranoia.gemspec
```diff
@@ -24,7 +24,7 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = '>= 2.7'
 
-  s.add_dependency 'activerecord', '>= 6', '< 8.1'
+  s.add_dependency 'activerecord', '>= 6', '< 8.2'
```
JulienDefrance, line 27: Should we instead have this be open-ended, to avoid similar complications whenever Rails 8.2 or 8.3 comes out?

rmm5t: No, not until there's something to add to the build matrix.

